PR #5, opened 2015-07-11
Hello world

adelcambre:
image

A first pass at a first bash exercise! I'm using [`bats`](https://github.com/sstephenson/bats) for the test suite which seems to work well (and is available in homebrew).


kytrinyx:
Merged with --ff in master. Thanks!

PR #6, opened 2015-07-11
Hamming

adelcambre:
Add the hamming exercise. Tests lifted directly from the ruby one. 

I'm not sure my example is the best (temp files, ewww), but it passes the tests.


kytrinyx:
W00t! Squashed, rebased, and merged in 3d1d0e69ee0e703554bfac896f957e9e94e32dd0

PR #10, opened 2016-03-16
Make the tests pass even if the scripts are not executable

kenden:
Execute the scripts to test with bash
(ex: 'bash script_name.sh')
instead of executing the scripts directly (ex: './script_name.sh'), which requires the files to be executable.

Else a user might be confused as to why the scripts don't pass the test.

I'm not 100% sure this is desirable behavior (it makes a shebang `#!/bin/bash` useless in the scripts), so please review.

Another approach might be to add more information in the readme file, like:

"Make sure the script is executable with `chmod +x script_name.sh`"

### DIFF
--- a/hamming/hamming_test.sh
+++ b/hamming/hamming_test.sh
@@ -1,104 +1,104 @@
 @test "Short identical strands have hamming distance of zero" {
-  run ./hamming.sh C C
+  run bash hamming.sh C C
 
   [ "$status" -eq 0 ]
   [ "$output" -eq 0 ]
 }
 
 @test "Long identical strands have hamming distance of zero" {
-  run ./hamming.sh CATSCATS CATSCATS
+  run bash hamming.sh CATSCATS CATSCATS
 
   [ "$status" -eq 0 ]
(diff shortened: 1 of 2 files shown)


kenden:
Fixes https://github.com/exercism/xbash/issues/9

hamming/hamming_test.sh
```diff
@@ -1,103 +1,103 @@
...
   [ "$status" -eq 1 ]
   [ "$output" = "The two strands must have the same length." ]
 }
 
 @test "no input" {
-  run ./hamming.sh
+  run bash hamming.sh
 
   [ "$status" -eq 1 ]
   [ "$output" = "Usage: ./hamming.sh <string1> <string2>" ]
```
adelcambre: This test fails now with the different invocation. Can you fix that up?

kenden:
@adelcambre Fixed, sorry about that.

adelcambre:
Thanks! Looks good to me.

kotp (approved):

PR #11, opened 2016-03-16
Fix hamming test containing invalid DNA letters

kenden:
Also added a couple of tests to avoid this.

### DIFF
--- a/exercises/hamming/hamming_test.sh
+++ b/exercises/hamming/hamming_test.sh
@@ -6,7 +6,7 @@
 }
 
 @test "Long identical strands have hamming distance of zero" {
-  run bash hamming.sh CATSCATS CATSCATS
+  run bash hamming.sh CATGCATG CATGCATG
 
   [ "$status" -eq 0 ]
   [ "$output" -eq 0 ]


adelcambre:
Can you add the code that makes this pass to the example? Thanks!

kenden:
@adelcambre I added code to make the example.sh pass the extra tests

kenden:
@adelcambre I am having second thoughts about adding new tests and new code to the example:
This make the exercise quite a bit more complicated. The examples in Go or Python do not have this checks.
I think I will simply replace this:
 @test "Long identical strands have hamming distance of zero" {
 -  run bash hamming.sh CATSCATS CATSCATS
 +  run bash hamming.sh CATGCATG CATGCATG
to not have some invalid DNA in there (just because it doesn't look good, nothing else).
And that's it.
What do you think?

kytrinyx:
@kenden I think that making the replacement you suggest sounds good.

kenden:
@kytrinyx I replaced the extra tests and extra code by just a little cleanup.

kytrinyx:
Thanks!

PR #12, opened 2016-03-16
Add bash leap exercise

kenden:
Added bash leap exercise.


kenden:
Updated the test to match master (no need to set the script as executable)

kenden:
Rebased on master (gigasecond and rna-transcription exercices were added), I had to merge config.json.

kytrinyx:
Sweet! Thanks so much.

exercises/leap/example.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
+
+USAGE="Usage: leap.sh <year>"
+
+# check number of params
+if  [ "$#" != 1 ]; then
```
0xmohit: It would have been good to turn usage into a function, like so:

``` sh
usage() {
    echo >&2 "Usage: leap.sh <year>"
    exit 1
}
```

and use the function where required.

kenden: @0xmohit
Thanks, that's better indeed. I made another PR to improve that.
https://github.com/exercism/xbash/pull/23

PR #17, opened 2016-05-22
Updating the Bash documentation page

cstackpole:
Many of the pages on the Bash page (http://exercism.io/languages/bash) are missing. They have helpful hints on what needs to be corrected/added. These commits address those helpful hints as a start point.


docs/INSTALLATION.md
```diff
@@ -0,0 +1,3 @@
+# Installation
```
kytrinyx: The docs shouldn't have titles, since the website adds section titles (and this will result in double titles).

kytrinyx:
Thanks!

I have just one comment about the format of the docs themselves (see inline).

The build is breaking on the `configlet .` command.
I think the fix is going to be to put `docs` into the `"ignored"` section of the `config.json` file. 
If you want to run `configlet` locally, you can first run `bin/fetch-configlet` and then `configlet .` (notice the dot) from the xbash directory.

cstackpole:
Thanks for the feedback. Just verified that configlet works. Cleaning up the headers now.

kytrinyx:
Thanks!

PR #20, opened 2016-07-23
Added Gigasecond and Rna Transcription exercises.

zwebb:
Added the Gigasecond and RNA Transcription exercises, copying the test cases from the Ruby versions of these problems.


kytrinyx:
Nice, thank you! ✨

PR #21, opened 2016-08-09
Add raindrops exercise

kenden:
Based on Ruby, Php, Rust versions

### DIFF
--- a/config.json
+++ b/config.json
@@ -9,7 +9,8 @@
     "gigasecond",
     "leap",
     "hamming",
-    "rna-transcription"
+    "rna-transcription",
+    "raindrops"
   ],
   "deprecated": [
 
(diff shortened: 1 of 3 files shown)


kytrinyx:
Nice, thank you!

kytrinyx:
I don't know if this is documented anywhere yet, but some exercises also have a JSON file with canonical data, which can sometimes make it easier to implement.

If you go to this URL http://x.exercism.io/v3/tracks/bash/todo (you'll want a json plugin for the browser), then look for the "data" key. If it's null you're out of luck, but if it has a URL, that should link to a file with test inputs/outputs that we've extracted from existing implementations.

kenden:
@kytrinyx thanks!
I'll take a look. I think a few bash exercises already needs fixing from this file :)

kytrinyx:
Haha, that's what I get for being so behind on documenting things.

Thank you :)

PR #23, opened 2016-08-10
Better leap example: DRY up displaying error

kenden:
Make usage into a function
This outputs to stderr and avoid repeating exit(1)


kytrinyx:
That's lovely, thanks!

PR #28, opened 2016-09-19
Fix bob bug: [[:space:]] does not match \u00a0 (on linux)

kenden:
Fix for https://github.com/exercism/xbash/issues/27

[[:space:]] does not match \u00a0 on linux.
I commented out the test and added a comment for now.

Also DRYed up the code a bit.

### DIFF
--- a/exercises/bob/example.sh
+++ b/exercises/bob/example.sh
@@ -6,6 +6,7 @@ input="$(printf %b "${1}")"
 
 # Remove space characters
 input="${input//[[:space:]]/}"
+input="${input//$(printf '\u00a0')}"  # [[:space:]] doesn't match \u00a0 on Linux
 
 # Is there silence?
 if [[ "${input}" == "" ]]; then
--- a/exercises/bob/example.sh
+++ b/exercises/bob/example.sh
@@ -1,29 +1,53 @@
 #!/bin/bash
 
+# Exit on error. Append || true if you expect an error.
+set -o errexit
+# Exit on error inside any functions or subshells.
+set -o errtrace
+# Do not allow use of undefined vars. Use ${VAR:-} to use an undefined VAR
+set -o nounset
+# Catch error even if it's not in after the last pipe.
+set -o pipefail
+# Turn on traces, useful while debugging but commented out by default
+# set -o xtrace
+
+# Set magic variables for current file, directory, os, etc.
+#__dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+#__file="${__dir}/$(basename "${BASH_SOURCE[0]}")"
+#__base="$(basename ${__file} .sh)"
+
+# print argument and exit
+function say() {
+  local mesg=${1}
+  echo ${mesg}
+  exit 0
+}
+
+if [[ $# -eq 0 ]]; then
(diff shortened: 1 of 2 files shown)


qjd2413:
I don't think this is the best solution to this-- it forces Linux users to either look at this solution or debug the testing suite. If `\u00a0` isn't matched by `[[:space:]]` on Linux or with another general solution, I think it should be removed from the test.

kenden:
Hmmm.
I used https://github.com/exercism/x-common/blob/master/exercises/bob/canonical-data.json
to create the exercises. 
In the future we might want to just import this file and runs tests from it.

As seen [here](https://github.com/exercism/x-common/pull/50) ( the PR where this test data was added), if exercises become unexpectedly hard to implement:

> A recourse (...) would be to generate code for all cases where "description" didn't equal "other whitespace".

So yes, skipping this test should be the proper approach. It's not very future-proof, though.
@kytrinyx do you agree?

Other possibilities: 
- remove this test from canonical-data.json
- leave it in and mark this exercises' difficulty much higher

qjd2413:
I think the test is useful-- The best course of action in my opinion is to remove "\u00a0" from it.

exercises/bob/example.sh
```diff
@@ -1,28 +1,53 @@
...
+  say "Fine. Be that way!"
+fi
+
 # Get what Bob says.
 # %b interprets escapted characters (like '\n' for newline)
 input="$(printf %b "${1}")"
 
 # Remove space characters
 input="${input//[[:space:]]/}"
+input="${input//$(printf '\u00a0')}" # [[:space:]] doesn't match \u00a0 on Linux
```
kotp: That is because that is a "no break space" not your normal space.

kenden: The strange thing is that it is working on mac (i.e. \u00a0 is considered a space), but not on linux.

kytrinyx:
@kenden @kotp @qjd2413 would you see if you can come to a conclusion about how to resolve this? I'd love to get this closed out.

kotp:
@kenden @qjd2413 Did we want to comment out the test with a note to explore?

Smarticles101:
I like @kotp's idea

kenden:
Ok, I took commented out the test and added a comment for now.

kytrinyx:
@kenden random side note... would you be interested in joining the maintainer team for this track? I appreciate the input that you've contributed here. As you probably noticed it's a pretty low-volume repository, and we wouldn't expect anything more from you than you're already doing unless you wanted to, aside from wielding merge powers when appropriate.

Smarticles101:
Thanks for all the hard work @kenden. It would be pretty cool if you accepted @kytrinyx's offer to become a maintainer. The bash track is pretty slow and some extra people to review and merge pull requests is always good to have

kenden:
@kytrinyx @Smarticles101 Sure, I'd be glad to help!

kytrinyx:
Yay, fantastic. I've sent you an invitation.

We have some (possibly woefully inadequate) documentation about some of the things that we typically think about when maintaining tracks here: https://github.com/exercism/docs/blob/master/maintaining-a-track/README.md - but like I said: you're welcome to do as much or as little as works for you!

PR #33, opened 2016-11-13
Add shebang for bats files and change file permission for simple run tests

platinumthinker:
I added shebangs for simple run bats files


kytrinyx:
Nice, thanks!

PR #37, opened 2017-01-18
dibs: I will implement exercise word-count

t-edward:


kotp:
I think the `config.son` needs a change @t-edward .  Thanks for jumping on this, by the way!

kotp:
Thank you @t-edward for the work on this.  Bringing in your changes hopefully soon!

PR #39, opened 2017-04-05
probot/stale requires the presence of a .github/stale.yml file.

kotp:
If you want the default behavior, you need to add an empty .github/stale.yml file to this repo, otherwise the bot will not be active.


kotp:
Closing without prejudice.  When we want it it can come in.

PR #42, opened 2017-06-08
dibs: I will implement exercise difference-of-squares

rpalo:
Trying to implement this exercise to get Bash one step closer to getting launched!  This is my first pull request ever ever, so please let me know if I do anything wrong. 😄

PR #44, opened 2017-06-10
add difference-of-squares

jaggededgedjustice:


kytrinyx:
Hi @jaggededgedjustice—this exercise was implemented a few days ago by @rpalo in #42 

Thanks so much for contributing! If you see another exercise you'd like to implement here, we'd love to see an implementation: http://exercism.io/languages/bash/todo

💜 🌷 💜

Issue #47, opened 2017-06-11
Error Handling exercise planning


jaggededgedjustice:
What should the error handling exercise cover for bash? The current list I can think of would be:
* Argument validation
* set -e
* set -u

Are there others?

Though I’m not sure how to test usage of the set flags . Perhaps the argument validation would be enough, that could resolve the remaining concerns from #29.

kotp:
Specifically for the Hello World exercise or a new exercise?

jaggededgedjustice:
New exercise, https://github.com/exercism/x-common/blob/master/exercises/error-handling/description.md

kotp:
In that case, definitely pile them up.

Does `getopts` give you the information for the set flags?

vlisivka:
Typical "safe" bash script uses `set -ue`, sometimes `set -ue -o pipefail`. I wrote whole library for bash which works with `set -ue` options.

jaggededgedjustice:
It turns out the $SHELLOPTS variable will contain values for the things controlled via set. So `echo $SHELLOPTS | grep ...` can tell you if a specific flag was used.

jaggededgedjustice:
I added #49. It doesn't do anything with 'set', the script being tested is run in a sub shell and the 'set -eu' options aren't inherited and don't escape a subshell. So I can't see a way to write a test and verify that a given option was used by the script. Also, bats is doing a 'set -e' which will be visible if you're trying to check SHELLOPTS after running your script.

vlisivka:
+jaggededgedjustice Why not to use `source` instead of `run` in this particular case?

jaggededgedjustice:
I had been thinking of having separate steps for each thing, so 1 function for arg checking, a different function to demonstrate use of 'set -e' and so on. But I can't see how to make that work, so maybe it's just 1 script and then it can be sourced to check the use of set flags. Depends if there are other situations to address besides arg checking.

kytrinyx:
Closed by #49

PR #49, opened 2017-06-17
Error handling exersise #47

jaggededgedjustice:
Currently only has argument checking

### DIFF
--- a/config.json
+++ b/config.json
@@ -141,6 +141,16 @@
         "string_comparison",
         "control_flow_conditionals"
       ]
+    },
+    {
+      "uudi": "b2b4ee35-108f-45ce-b294-c10d332e5579",
+      "slug": "error-handling",
+      "core": true,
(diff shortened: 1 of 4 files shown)


kotp:
So far so good.

jaggededgedjustice:
Reworked so the script can be sourced to check that `set -eu` has been used

kytrinyx:
Thanks @jaggededgedjustice! If you add the exercise to the `config.json` file's "exercises" array, we should be able to merge the pull request.

kytrinyx:
@jaggededgedjustice Thanks! Would you rebase against the most recent upstream master? We've added some new keys that you'll need to add (if you get stuck I can go ahead and do this for you, but I wanted to give you a chance to do it if you have time).

kytrinyx:
This looks great, @jaggededgedjustice. Once the build passes I'll get this merged. 🌷 Thank you

PR #52, opened 2017-07-08
two-fer: add to track

Smarticles101:
Add two-fer to track and update hello-world to match canonical data


kytrinyx:
I've rebased this onto the most recent master, and will merge it as soon as it passes CI. Thanks @Smarticles101 ✨

PR #76, opened 2017-08-12
Introducing Bash

kotp:
About Bash.  Just a blurb, can be expanded.

exercism/meta#85


m-a-ge:
This is not really related to the PR itself, but it's still worth mentioning here.

I've checked why Travis CI build failed by looking through logs and discovered the following:

- `-> The implementation for 'word-count' is missing a test suite.`

I've created #77 for it

Smarticles101:
This should be able to merge after a rebase so the build passes @kotp

PR #80, opened 2017-08-26
implement exercise nucleotide-count

Smarticles101:
I did my best to choose topics and a difficulty for the exercise. The only last two things are should it be a core exercise? If not, what should it be unlocked by?

### DIFF
--- a/exercises/nucleotide-count/README.md
+++ b/exercises/nucleotide-count/README.md
@@ -1,0 +1,38 @@
+# Nucleotide Count
+
+Given a DNA string, compute how many times each nucleotide occurs in the string.
+
+DNA is represented by an alphabet of the following symbols: 'A', 'C',
+'G', and 'T'.
+
+Each symbol represents a nucleotide, which is a fancy name for the
+particular molecules that happen to make up a large part of DNA.
+
+Shortest intro to biochemistry EVAR:
+
+- twigs are to birds nests as
+- nucleotides are to DNA and RNA as
+- amino acids are to proteins as
+- sugar is to starch as
+- oh crap lipids
+
+I'm not going to talk about lipids because they're crazy complex.
+
+So back to nucleotides.
+
+DNA contains four types of them: adenine (`A`), cytosine (`C`), guanine
+(`G`), and thymine (`T`).
+
+RNA contains a slightly different set of nucleotides, but we don't care
+about that for now.
+
+Run the tests with:
+
+    bats whatever_test.sh
+
+## Source
+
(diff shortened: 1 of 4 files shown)


kotp:
This looks good to me.  I did notice that there are no tests that are skipped.  @exercism/bash are we skipping all but the first test by default to reduce the noise when solving these, until we pass each test?

Smarticles101:
@kotp It doesn't look like any of the other exercises do this, but it might be a good policy to adapt.

kotp:
For the Ruby and Delphi tracks we skip all but the first test.  The very first exercise, I think for both tracks, have comments with an explanation of what to do to unskip them, and perhaps also duplicated in the Readme. 

This could definitely be a different pull request, that could touch the existing exercises.

Smarticles101:
Yeah, that's what we do in the java track too. I'll open another issue for that and start working on it. In the meantime, should we decided on core exercise/unlocked by?

kotp:
Leap, Bob, and a couple others are not unlocked by Pangram.  But this being unlocked by Pangram seems to be OK, at least at first thought.

exercises/nucleotide-count/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+DNA contains four types of them: adenine (`A`), cytosine (`C`), guanine
+(`G`), and thymine (`T`).
+
+RNA contains a slightly different set of nucleotides, but we don't care
+about that for now.
+
+Run the tests with:
+
+    bats whatever_test.sh
```
kenden: replace with `bats nucleotide_count_test.sh` ?

Smarticles101: The readme file for every exercise is generated by configlet. This is the track-specific insert that we have for this track. (EXERCISE_README_INSERT](https://github.com/exercism/bash/blob/master/docs/EXERCISE_README_INSERT.md)

exercises/nucleotide-count/example.sh
```diff
@@ -0,0 +1,38 @@
...
+	exit
+}
+
+if [ ! -n "$input" ]; then
+	end
+fi
+
+for i in $(seq 1 ${#input}); do
+	char=${input:i-1:1}
+	if [ $char = "A" ]; then
```
kenden: [shellcheck](https://github.com/koalaman/shellcheck) says:
"Double quote to prevent globbing and word splitting."
--> replace $char by "$char"

Also, to have a better standard in the example, replace $char with ${char} ? 
See [this](https://stackoverflow.com/questions/8748831/when-do-we-need-curly-braces-around-shell-variables) for ex.

So `$char -> "${char}"`

kenden:
Thanks for the PR! LGTM

PR #85, opened 2017-09-07
Add skips to tests

### LINKED ISSUE #82: Add ignore's to every test but the first for every exercise
In each exercise, every test should be ignored except the first one. Documentation about this should be provided in the first exercise Ref. #80

Smarticles101:
Add skips to tests, and a comment about them. Closes #82

### DIFF
--- a/exercises/bob/bob_test.sh
+++ b/exercises/bob/bob_test.sh
@@ -8,6 +8,7 @@
 }
 
 @test "shouting" {
+  skip
   run bash bob.sh 'WATCH OUT!'
 
   [ "$status" -eq 0 ]
@@ -15,6 +16,7 @@
 }
 
 @test "shouting gibberish" {
+  skip
   run bash bob.sh 'FCECDFCAAB'
 
   [ "$status" -eq 0 ]
@@ -23,6 +25,7 @@
 
 
 @test "asking a question" {
+  skip
   run bash bob.sh "Does this cryogenic chamber make me look fat?"
 
   [ "$status" -eq 0 ]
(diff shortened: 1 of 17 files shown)


exercises/gigasecond/gigasecond_test.sh
```diff
@@ -8,27 +8,35 @@
 }
 
 @test 'June 13th, 1977' {
+  # this is used to skip the test
+  # normally, we skip every test except for the first one
+  # this allows for a person to focus on solving a test at a time
+  # you can comment out or delete the `skip` to run the test when ready
+  skip
```
kotp: Comment out the first skip as an example, then they can see how they should modify the next tests to emulate the behavior of that first test.  Leave the suggestion to delete the skip as well, though, as that is definitely another option.

kotp: Perhaps do the same for every exercise (minus the comment) which can serve as a reminder in case a student leaves the track for a time, and comes back, as they get the reminder.

Smarticles101: sounds like a good idea 👍

exercises/bob/bob_test.sh
```diff
@@ -217,6 +219,16 @@
...
+=======
+@test "other whitespace" {
+  skip
+  run bash bob.sh "\n\r \t\u000b\u00a0\u2002"
+
+  [ "$status" -eq 0 ]
+  [ "$output" = "Fine. Be that way!" ]
+}
+
+>>>>>>> e1ed45efbe5be55ce35968443e5d4b454adb1935
```
kotp: Looks like some merge commit markings are here?

Smarticles101:
Seeing as the comments made by @kotp on October 7th have been resolved, I feel as though this is appropriate to merge now

PR #93, opened 2017-10-11
Add an acronym exercise for the bash language in exercism.io

netserf:
Adding my first programming exercise to exercism.io.

### DIFF
--- a/config.json
+++ b/config.json
@@ -164,15 +164,6 @@
     },
     {
       "core": false,
-      "difficulty": 1,
-      "slug": "acronym",
-      "topics": [
-        "string_transformation",
-        "input_validation"
-      ],
-      "uuid": "5812c10a-aee1-11e7-add7-ef139384f6eb"
-    },
-      "core": false,
       "difficulty": 5,
       "slug": "atbash-cipher",
       "topics": [
@@ -182,7 +173,17 @@
       ],
       "unlocked_by": "hamming",
       "uuid": "03808072-0228-6580-7a9b-cc709a8a5b8c0e600b9"
-    }
+    },
+    {
+      "core": false,
+      "difficulty": 1,
+      "slug": "acronym",
+      "topics": [
+        "string_transformation",
+        "input_validation"
+      ],
+      "uuid": "5812c10a-aee1-11e7-add7-ef139384f6eb"
+    }    
   ],
   "foregone": [],
   "language": "Bash"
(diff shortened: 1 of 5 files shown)


Smarticles101:
Looks good! the build failed because you haven't put acronym into the config.json file. If you have any questions, just ask!

netserf:
Sorry about that and thanks for the tip. I'll take a look.

netserf:
Okay, I have a few things to learn about git still. I hope this is better now.

config.json
```diff
@@ -173,8 +173,18 @@
       ],
       "unlocked_by": "hamming",
       "uuid": "03808072-0228-6580-7a9b-cc709a8a5b8c0e600b9"
-    }
+    },
+    {
+      "core": false,
```
NobbZ: If it is not `core`, it needs to be `unlocked_by` something.

netserf: Thank you for the feedback Nobbz. Could you point me to the docs on config.json? I guessed at the requirements, but I should really understand it if I'm to add more exercises in the future.

NobbZ: Well, I was told today that this is in fact not an issue. If the key is missing, it is implied to be unlocked by the very first exercise of the track.

More info: exercism/configlet#96

netserf: I read the [exercism/configlet#96](https://github.com/exercism/configlet/issues/96). I can't say I understand the requirements of config.json completely, but as long as I've met the minimum requirements that should work for now. If this isn't documented anywhere, where would be the best place be to add the docs?

Smarticles101:
@netserf Do you want to estimate a difficulty and position in the track along with a good choice for it to be `unlocked_by` or not if you don't see an obvious exercise to unlock it. If not, I can open an issue for whichever of these (or both) you decide not to do.

Also, recently we've added `skips` to exercise test suites. Could you add a `skip` to the first line of every test function with the first one commented out?

Sorry for such a late response 😬  and thanks for the hard work 👍

netserf:
@Smarticles101, sorry it took me a bit to circle back around to this. 
- Estimate a difficulty 
  Calling it a 1 (i.e. not very difficult)
- Position in the track
  Is this the relative position in the config.json file? I'll put it after leap. Please let me know if I'm not understanding this correctly.
- unlocked_by
  Done
- add skip to the first line of every test function with the 1st one commented out
  Done

Smarticles101:
@netserf yep, that's exactly what I was looking for, thanks! Sorry it took me a few days to get back to this. :)

netserf:
Thanks for walking me through the process.  :-)

PR #96, opened 2017-10-15
Generate CONTRIBUTING.md with specific step-by-step instructions

rpalo:
I made a thing.  I found myself having to bounce around to a bunch of different places in the Exercism documentation while I was porting exercises over, so I wrote the whole process (including specifics to the Bash track) down.  I'm not sure you can use it, or if the CONTRIBUTING.md document is the right place for it, but let me know what you think.  I'm happy to make as many changes/edits as you like.

I also know that having documentation duplicated, like this duplicates those separate existing docs, causes a more maintenance as things change.  I'm happy to keep an eye on that and update if need be.


Smarticles101:
Sorry for the late response, still looks good to me so I'll merge it. Thanks @rpalo contributing docs are great to have and hopefully this will get more people comfortable with contributing to the bash track :)

PR #101, opened 2017-12-19
reorder config by difficulty

sjwarner-bp:
I noticed that the final exercise was only rated as a 2 - maybe it was just appended?

Re-ordered in difficulty order.

### DIFF
--- a/config.json
+++ b/config.json
@@ -164,18 +164,6 @@
     },
     {
       "core": false,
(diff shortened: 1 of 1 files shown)


Smarticles101:
@sjwarner-bp Yeah, I'm not sure how that got there, thanks for the quick fix

PR #106, opened 2017-12-27
fix contributing highlighting 'invalid' json

sjwarner-bp:
I've removed the `json` identifier from the json code block in `CONTRIBUTING.md`, as it was being highlighted in red (invalid) as json does not support comments.

Personally, I think this is slightly ugly looking, and as it isn't *really* json, I think it is an improvement to remove the red highlighting.

Let me know what the rest of you think and if this is seen as an improvement we can merge it :slightly_smiling_face:


budmc29:
What about using ```javascript? This will give the syntax highlighting as the original author wanted.

sjwarner-bp:
I wasn't too sure what would be best - I'll change it to javascript 🙂

PR #107, opened 2017-12-27
grains: add to track

sjwarner-bp:
Add `grains` to track

### DIFF
--- a/exercises/grains/grains_test.sh
+++ b/exercises/grains/grains_test.sh
@@ -61,6 +61,7 @@
   run bash grains.sh 0
 
   [ "$status" -eq 1 ]
+  [ "$output" = "Error: invalid input" ]
 }
 
 @test "Square -1 causes error" {
@@ -68,6 +69,7 @@
   run bash grains.sh -1
 
   [ "$status" -eq 1 ]
+  [ "$output" = "Error: invalid input" ]
 }
 
 @test "Square 65 causes error" {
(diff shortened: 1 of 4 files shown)


exercises/grains/grains_test.sh
```diff
@@ -0,0 +1,78 @@
...
+@test "Square 1" {
+  #skip
+  run bash grains.sh 1
+
+  [ "$status" -eq 0 ]
+  [ "$output" = "1" ]
+}
+
+@test "Square 2" {
+  #skip
```
budmc29: Skips are commented.

exercises/grains/grains_test.sh
```diff
@@ -0,0 +1,78 @@
...
+  run bash grains.sh -1
+
+  [ "$status" -eq 1 ]
+}
+
+@test "Square 65 causes error" {
+  #skip
+  run bash grains.sh 65
+
+  [ "$status" -eq 1 ]
```
budmc29: Expects error.

exercises/grains/grains_test.sh
sjwarner-bp: Oops! 🙂

exercises/grains/grains_test.sh
sjwarner-bp: Would you agree the error should be tested by something like:

budmc29: Yes, that looks good.

sjwarner-bp:
Hopefully this is better 🙂

budmc29:
Thank you.

PR #108, opened 2017-12-28
collatz-conjecture: add to track

sjwarner-bp:
Add to track as part of fleshing out track, in preparation for launch.

I've not put the logic in a function - I figured that this should be ok, but let me know what you think!

### DIFF
--- a/exercises/collatz-conjecture/collatz_test.sh
+++ b/exercises/collatz-conjecture/collatz_test.sh
@@ -1,0 +1,47 @@
+#!/usr/bin/env bats
+
+@test "zero steps for one" {
+  #skip
+  run bash collatz.sh 1
+
+  [ "$status" -eq 0 ]
+  [ "$output" -eq 0 ]
+}
+
+@test "divide if even" {
+  skip
+  run bash collatz.sh 16
+
+  [ "$status" -eq 0 ]
+  [ "$output" -eq 4 ]
+}
+
+@test "even and odd steps" {
+  skip
+  run bash collatz.sh 12
+
+  [ "$status" -eq 0 ]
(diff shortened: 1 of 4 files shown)


exercises/collatz-conjecture/collatz_test.sh
```diff
@@ -0,0 +1,47 @@
...
+
+  [ "$status" -eq 0 ]
+  [ "$output" -eq 152 ]
+}
+
+@test "zero is an error" {
+  skip
+  run bash collatz.sh 0
+
+  [ "$status" -eq 1 ]
```
budmc29: This expects an error.

exercises/collatz-conjecture/collatz_test.sh
```diff
@@ -0,0 +1,47 @@
...
+}
+
+@test "negative value is an error" {
+  skip
+  run bash collatz.sh -15
+
+  [ "$status" -eq 1 ]
```
budmc29: Expects error as well.

exercises/collatz-conjecture/collatz_test.sh
sjwarner-bp: Would you agree the error should be tested by something like:

```
[ "$output" -eq "Error: invalid input" ]
```

and that the error should be echo'ed out? 🙂

sjwarner-bp:
Hopefully this is more like what you were expecting 🙂

budmc29:
Thank you.

exercises/collatz-conjecture/collatz_test.sh
budmc29: Yes.

PR #110, opened 2017-12-29
luhn: add to track

sjwarner-bp:
As part of fleshing out the track, I've added `luhn`.

I need to fix the conflict, but once someone has approved it then I'll fix that in the web-editor and merge in 🙂

### DIFF
--- a/config.json
+++ b/config.json
@@ -164,6 +164,19 @@
     },
     {
       "core": false,
+      "difficulty": 3,
+      "slug": "luhn",
+      "topics": [
+        "algorithms",
+        "control_flow_conditionals",
+        "integers",
+        "mathematics"
+      ],
+      "unlocked_by": "difference-of-squares",
+      "uuid": "a3b16d02-2963-445d-9339-7e3002375a71"
+    },
+    {
+      "core": false,
       "difficulty": 5,
       "slug": "atbash-cipher",
       "topics": [
--- a/exercises/luhn/README.md
+++ b/exercises/luhn/README.md
@@ -1,0 +1,76 @@
+# Luhn
+
+Given a number determine whether or not it is valid per the Luhn formula.
+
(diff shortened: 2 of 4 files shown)


budmc29:
Looks good, can you uncomment all the tests except the first one?

sjwarner-bp:
Oops, silly mistake from testing! On it now 👍 

Thanks for the reviews 🙂

sjwarner-bp:
Should be fixed - once it's approved I'll fix the conflict and then merge 🙂

budmc29:
The problem-specification for this exercise asks that the expected output is either true or false, what do you say about replacing the exit status with an explicit return or print in the tests?

Bash exit code of 0 means true, but other languages treat 0 as false and 1 as true, and this might cause confusion, and at the same time allows users to use the script from the CLI and see an output which makes the default use case for bash scripts.

I have nothing against integrating the exercise as it is, but it might be worth discussing about this since looking at the other exercises all of them except `armstrong_numbers` and `triangle` are following this convention.

sjwarner-bp:
That makes sense to me - I can get working on it, and once that's done correct the other two!

It would be a good idea for us to document this somewhere (alongside justification) 🙂

budmc29:
Added info with https://github.com/exercism/bash/pull/123.

sjwarner-bp:
Think this should do it - let me know what you think @budmc29 🙂

budmc29:
Maybe is better to output to the terminal `true` just like the specification, it will help avoiding the confusion for 0 being true or false.

sjwarner-bp:
Like this? 🙂

budmc29:
Awesome, that look very good. Thank you.

budmc29:
I'm ready to merge it if you resolve the conflict.

sjwarner-bp:
Perfect, I'll resolve the conflict when I'm back at my laptop! 🙂

PR #113, opened 2017-12-29
add MAINTAINING.md doc

sjwarner-bp:
Add markdown to describe the duties of maintaining a repo.

This is heavily influenced by the Java track's document, as I think it is a great example.


budmc29:
Looks good.

PR #121, opened 2017-12-30
_template: add a template for quickly porting exercises to track

sjwarner-bp:
Adds a `_template` directory with a basic exercise outline, to facilitate more uniform contributions.

### DIFF
--- a/_template/example.sh
+++ b/_template/example.sh
@@ -1,3 +1,3 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # A solution to the exercise, passing all tests, should be here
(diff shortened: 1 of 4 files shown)


_template/example.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
```
budmc29: Can we use `#!/usr/bin/env bash` here?

sjwarner-bp: Sure thing 🙂 

There are a good few exercises that use the old shebang though - I'll try and work through correcting them, if we think this should be a track-wide change?

kotp:
I think it is good to go!

_template/example.sh
budmc29: Great idea.

PR #123, opened 2017-12-30
Add test output check suggestion

budmc29:
Check for actual return value instead of bash exit code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -61,11 +61,11 @@ Currently, you're interested in all of the `"cases"`.  Each of these is a test c
 Implement each of the test cases (along with any comments or explanations you think are necessary for people who aren't sure of themselves).  When in doubt, check out the other, already implemented exercises for an example.  Here's an example of implementing the test case shown above.
 
 ```bash
-#!/bin/bash
+#!/usr/bin/env bats
 
 @test 'Say Hi!' {
   run bash hello_world.sh
-  
+
   [ "$status" -eq 0 ]
   [ "$output" = "Hello, World" ]
 }
@@ -75,6 +75,9 @@ Implement each of the test cases (along with any comments or explanations you th
 
 For each test you create, you should use the `description` value for the test case description.  The `property` value is generally used by other languages to specify what the function to be run is called.  Use your best judgement and check out how other exercises handle this.  If there's just one property, you can just test the script as a whole.  If there's multiple properties, consider testing those as either subcommands or flags (e.g. an exercise with an encode and a decode property could be handled by `run bash cyper.sh encode <argument>` or by `run bash cypher.sh -e <argument>`).  You're the one writing the tests, so you decide.  You'll get feedback when you open your pull request anyways, so don't stress too much about it.
 
+For test cases that require the implementation to return `true` or `false` make sure you are requiring the user to output a value, for example "true" or `true`, to the console instead of just relying on the exit status in bash (0 for success, 1-255 for error).
+This will be consistent with the existing exercises and prevent possible issues in the future.
+
 You can check to see if your example script works by running `bats`.
 
 ```bash
@@ -83,7 +86,7 @@ $ bats hello_world_test.sh
 
 ### Implementing an Example Solution
 
-If you've been following along so far, your tests should fail.  Go ahead and implement your solution in the `<your-exercise>.sh` file.  Make sure your file has `#!/bin/bash` at the top.
+If you've been following along so far, your tests should fail.  Go ahead and implement your solution in the `<your-exercise>.sh` file.  Make sure your file has `#!/usr/bin/env bash` ([shebang](https://en.wikipedia.org/wiki/Shebang_(Unix))) at the top.
 
 Keep running your test file against it until your tests all pass.  This process should help ensure that both your tests *and* your example are ship shape!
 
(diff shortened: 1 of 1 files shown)


kotp:
The root slash is missing in `#!bin/bash` and it is not guaranteed to be installed/available in that place, while `/usr/bin/env bash` will allow it to be found wherever it is located in the users path (the first, specifically).

This allows the user to dictate better which version is used, as they have control of their path.

sjwarner-bp:
Cool - makes sense, just wanted to check 🙂 Thanks!

CONTRIBUTING.md
```diff
@@ -75,6 +75,9 @@ Implement each of the test cases (along with any comments or explanations you th
 
 For each test you create, you should use the `description` value for the test case description.  The `property` value is generally used by other languages to specify what the function to be run is called.  Use your best judgement and check out how other exercises handle this.  If there's just one property, you can just test the script as a whole.  If there's multiple properties, consider testing those as either subcommands or flags (e.g. an exercise with an encode and a decode property could be handled by `run bash cyper.sh encode <argument>` or by `run bash cypher.sh -e <argument>`).  You're the one writing the tests, so you decide.  You'll get feedback when you open your pull request anyways, so don't stress too much about it.
 
+For test cases that require the implementation to return `true` or `false` make sure you are requiring the user to output a value, for example "true" or `true`, to the console instead of just relying on the exit status in bash (0 for success, 1-255 for error).
```
kotp: This is a very long line...

CONTRIBUTING.md
```diff
@@ -75,6 +75,9 @@ Implement each of the test cases (along with any comments or explanations you th
...
+This will be consistent with the existing exercises and prevent possible issues in the future.
```
kotp: ... and this line is directly after the other line.  However, the idea of a paragraph is a blank line between (as indicated other places in this file).

So I am unsure if this is meant to be a new paragraph or was a mistake.

It is probably well known that I prefer around <= 80 characters per line,   I prefer paragraphs to be separated by an empty line.

budmc29: I'll fix those, thank you.

budmc29:
@kotp what do you think about the new wording and structure? I'm open to a better wording if you have any suggestions.

sjwarner-bp:
Build was failing for the same reason listed in #127.

Issue has since been resolved - restarting build now, it should pass.

CONTRIBUTING.md
```diff
@@ -83,7 +85,7 @@ $ bats hello_world_test.sh
 
 ### Implementing an Example Solution
 
-If you've been following along so far, your tests should fail.  Go ahead and implement your solution in the `<your-exercise>.sh` file.  Make sure your file has `#!/bin/bash` at the top.
+If you've been following along so far, your tests should fail.  Go ahead and implement your solution in the `<your-exercise>.sh` file.  Make sure your file has `#!/usr/bin/env bash` ([shebang](https://en.wikipedia.org/wiki/Shebang_(Unix))) at the top.
```
kotp: For readability, the link could go after this line as a note (or as a footnote, if that works with Github Markdown).

CONTRIBUTING.md
```diff
@@ -176,7 +178,7 @@ It's possible to submit an incomplete solution so you can see how others have co
 
 ### Cleaning Up
 
-20. If you've been practicing good version control throughout this process, you may have several commits.  At this point, you're almost ready to submit your pull request, but you should rebase against the most recent upstream master branch.
+If you've been practicing good version control throughout this process, you may have several commits.  At this point, you're almost ready to submit your pull request, but you should rebase against the most recent upstream master branch.
 
 ```bash
 # Assuming you've alread 'git added' and 'git commited'
```
kotp: This has a missing `y` in 'alread`.

CONTRIBUTING.md
```diff
@@ -188,7 +190,7 @@ $ git fetch upstream
 $ git rebase upstream/master
 ```
 
-21. Now you're ready to sync up with Github and open your pull request!
+Now you're ready to sync up with Github and open your pull request!
 
 ```bash
 $ git push --force-with-lease origin <your-branch-name>
```
kotp: It is not in the scope of this PR, but why suggest a force push at all... even with lease.  This should be a workflow management decision on the user, perhaps.

budmc29: I agree, but for now I think we can leave it as it is since it's a suggestion? Or do you think it will cause harm, I'm not aware of what `--force-with-lease` does (I did google it for a quick understanding).

CONTRIBUTING.md
budmc29: Yes, looks like it works.

CONTRIBUTING.md
```diff
@@ -75,6 +75,8 @@ Implement each of the test cases (along with any comments or explanations you th
 
 For each test you create, you should use the `description` value for the test case description.  The `property` value is generally used by other languages to specify what the function to be run is called.  Use your best judgement and check out how other exercises handle this.  If there's just one property, you can just test the script as a whole.  If there's multiple properties, consider testing those as either subcommands or flags (e.g. an exercise with an encode and a decode property could be handled by `run bash cyper.sh encode <argument>` or by `run bash cypher.sh -e <argument>`).  You're the one writing the tests, so you decide.  You'll get feedback when you open your pull request anyways, so don't stress too much about it.
 
+For test cases that are expected to return `true` or `false` make sure you are expecting an output value. It is better to return "true" or `true`, to the console instead of just relying on the exit status in bash (0 for success, 1-255 for error). This will be consistent with the existing exercises and prevent possible issues in the future.
```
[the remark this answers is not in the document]
budmc29: Fixed.

CONTRIBUTING.md
kotp: TL;DR: I think I wanted to have this flagged for discussion before I forgot about it.  It should be changed, though not in this PR.
- - -

It is advising on workflow, and this change surely doesn't belong to this PR, but I think the suggestion to use either form of 'force' is one that out-lies the scope of the Bash track.  If this is a suggestion that is applicable as a "This is how this Exercism does it" then it should be at a higher level.

The advice itself isn't bad advice, but, here, it is neglectfully incomplete, as it is not even close to a magic bullet solution, and can itself cause some problem.

budmc29:
Thank you.

PR #125, opened 2017-12-31
add #!/usr/bin/env bash shebang

sjwarner-bp:
As we move towards preferring the user's path to locate `bash` (#123, #121), I've updated the existing exercises with this change.

Let me know what you think about this 👍

### DIFF
--- a/exercises/acronym/example.sh
+++ b/exercises/acronym/example.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 if [ "$#" -ne 1 ]; then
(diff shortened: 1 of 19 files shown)


kotp:
Looks good to me.  We should be careful about including other changes that don't really fit the commit. (the couple of ending new lines in the files that were changed).

sjwarner-bp:
Oops, my mistake! I think my text editor must have stuck those in for some reason...

kotp:
We will let it slide this time. :see_no_evil: :hear_no_evil: :hear_no_evil: :stuck_out_tongue_winking_eye:

PR #127, opened 2018-01-02
Update error expectation to match specification

budmc29:


sjwarner-bp:
I've just pushed an update to add `triangle` back in. Hopefully this fixes it!

budmc29:
Thanks.

Issue #109, opened 2017-12-28
Core exercises


sjwarner-bp:
Looking through `config.json`, I have noticed we have few core exercises, but all exercises have an `unlocked_by`. I believe all 'optional' exercises need to be unlocked by core exercises (that's how it works on the Java track).

@exercism/bash Let me know if there is a reason we wouldn't have this policy, otherwise I'll adopt it and work on a fix.

kotp:
There definitely should be some basis of "introductory" and then the unlocked exercises from there.... I think this is a good adoption.

sjwarner-bp:
I am glad you agree @kotp - do you have any suggestions for which exercises should be "core"? I am going to be checking / amending the `config.json` as part of #103, and will be happy to tack this on afterwards 🙂

kotp:
Currently, it looks like:
* `hello-world`
* `leap`
* `error-handling`
* `pangram`

Are the current "core" exercises.

sjwarner-bp:
This is good to know - I'd suggest adding a couple more in, maybe a cipher based exercise and some arithmetic exercises. I think that:

* two-fer
* hamming
* armstrong-numbers
* atbash-cipher

are all good candidates to be added as core exercises. What does everyone else think @exercism/bash ?

budmc29:
I agree for `hamming`, `armstrong-numbers`, `atbash-cipher` to be added.

Maybe we should add `bob` as well, since it's a fun but challenging exercise in bash that involves RegEx.

We should focus on 7-8 core exercises to match a similar ration for core to non-core exercises like othertracks, and then focus on adding more exercises to the track.

budmc29:
Proposed core exercises:

1 `hello-world`
2 `leap`
3 `error-handling`
4 `pangram`
5 `hamming`
6 `armstrong-numbers`
7 `atbash-cipher`
8 `bob`

What do you think about these?

We can of course update them in the future, but at least we have a place to start.

sjwarner-bp:
Agreed. I'll update my PR (apologies for being slow with that - hectic week!)

sjwarner-bp:
Hi @budmc29. I am just looking over these, and I've realised that

`error-handling` cannot be both core AND `core: false`, `unlocked_by: none` as agreed in #128.

Currently, the core exercises stand as:

`grains`
`hello-world`
`reverse-string`
`armstrong-numbers`
`hamming`
`pangram`
`luhn`

[...]

Issue #103, opened 2017-12-26
Topics and difficulty


sjwarner-bp:
As per issues #62 - #74, a list of topics need to be added to `config.json` for each exercise.

Looking over `config.json`, it seems that most exercises have a minimum of 2 topics listed. Should these open issues now be closed, or are they left open for any particular reason? I think that these have all been resolved and would like to close them.

@Smarticles101 would you agree? I like to minimise clutter so we are able to focus on the matters at hand. Additionally, in #4 you mentioned the range of difficulty in exercises didn't seem right. Do you think we should amend this by reviewing the current difficulties, or just seek to implement some more difficult exercises? :slightly_smiling_face: 

Edit: regarding reviewing difficulties, both `two-fer` and `word-count` are rated 2 here, but the [`word-count`'s example implementation](https://github.com/exercism/bash/blob/master/exercises/word-count/example.sh) seems to be a bit more complex than [`two-fer`'s example](https://github.com/exercism/bash/blob/master/exercises/two-fer/example.sh).

sjwarner-bp:
I have just seen in #61 that Katrina has added some topics to each exercise, but would like someone more conversant in bash to review and amend.

I'll pick this up - this should de-clutter the issues a fair bit, allowing us to pick up some more discussion and move forward with launch plans! 🎉

kotp:
As of this moment, it appears that:

* `hello-world`
* `leap`
* `error-handling`
* `pangram`

are marked as "core", which should indicate that it should unlock certain exercises.

For reference here:

* `gigasecond` unlocked by ``
* `bob` unlocked by ``
* `acronym` unlocked by ``
* `raindrops` unlocked by `leap`
* `difference-of-squares` unlocked by `leap`
* `armstrong-numbers` unlocked by ``
* `anagram` unlocked by `pangram`
* `hamming` unlocked by `pangram`
* `rna-transcription` unlocked by `pangram`
* `two-fer` unlocked by `pangram`
* `nucleotide-count` unlocked by ``
* `word-count` unlocked by `pangram`
* `phone-number` unlocked by `pangram`
* `triangle` unlocked by `leap`
* `atbash-cipher` unlocked by `hamming`

sjwarner-bp:
That's great @kotp - thanks for doing that 🙂 

I think we are a bit too reliant on certain exercises. I've added a couple of suggestions in #109 which should diversify the track a little.

sjwarner-bp:
Closed via #128

Issue #75, opened 2017-07-22
We need a rough estimate of difficulty for each exercise


kytrinyx:
We've never really taken a pass through this track to figure out the relative difficulties of each exercise on a scale from 1 - 10 (1 == easy, 10 == hard).

If we could order the exercises so that the difficulty ramps up more gently, we can provide a better experience to those who are working through the exercises.

These are the exercises that we have

- hello-world
- gigasecond
- leap
- hamming
- rna-transcription
- raindrops
- word-count
- bob
- difference-of-squares
- anagram
- pangram
- two-fer
- phone-number

Note that hello-world was pretty difficult, and we've fixed this by simplifying hello world and adding a new exercise, two-fer, which is equivalent to the old one.

/cc @patbl, @dantiel, @mkrehbs, @deepbsd, @canel-rom1, @rpalo, @ConstantlyLost, @fwten, @fbi1714, @timmyjose, @mattj-io, @nielssorensen
You have all done at least some of these. Do you have any ideas about relative difficulties?

budmc29:
hello-world: easy
gigasecond: easy

fbi1714:
hello-world = 1
word-count = 8

rpalo:
I've got some suggestions for topics and difficulties.  Would it be better for me to leave comments on each of the relevant issues?  Or, I can put together a pull request with all of my suggested tweaks to `config.json` to be evaluated all in one place.  Whatever you think is best.

In the meantime:
hello-world = 1
gigasecond = 3
bob = 3
leap = 1
raindrops = 1
difference-of-squares = 1
pangram = 2
anagram = 4
hamming = 2
rna-transcription = 2
two-fer = 1
nucleotide-count = 2
word-count = 3
phone-number = 2
error-handling = 1
atbash-cipher = 5

I tried to pitch these numbers a little low, because I've seen other problems on other language tracks that haven't been ported over (yet 😀 ), and I know there's much harder problems yet to come.

Anyways, let me know if a pull request with updated/modified difficulties and topics would be useful or if I should get to work commenting on each separate issue 👍

budmc29:
> I've got some suggestions for topics and difficulties. Would it be better for me to leave comments on each of the relevant issues?

I think that's a very good idea if you have the time, it would help to explain why you gave each exercise a different rating.

> I can put together a pull request with all of my suggested tweaks to config.json to be evaluated all in one place

That can be done after we have more opinions about the difficulty of each exercise from various developers.

[...]

Issue #74, opened 2017-07-22
Topics: what is phone-number about?


kytrinyx:
In preparation for the launch of the Exercism redesign, we need to decide what topics to list for phone-number.
See https://github.com/exercism/bash/issues/61 for context.

What topics would you say that phone-number touches upon?

sjwarner-bp:
Closed via #128

PR #128, opened 2018-01-03
config.json: review exercise difficulty/unlocks/ordering

### LINKED ISSUE #61: We need topics for each exercise for the new prototype
We are working on a redesign of Exercism, which will have a much richer way of structuring the exercises.

The short version is that we'll have a handful of "core" exercises that are essential for learning about the language (typically 15-20), where completing each exercise unlocks a bunch of optional exercise that let you dive more deeply into different topics and practice different aspects of the language.

You can read more about this design choice in https://github.com/exercism/docs/blob/master/about/conception/progression.md

In order to make this possible, we need to figure out which topics each individual exercise covers. We have a [list of potential topics](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt) that we can draw from, but this is by no means an exhaustive list, and the topics might not be relevant to Bash at all.

As an experiment I'm creating a separate issue for each exercise so that I can ping people who have solved it to help figure this out.

sjwarner-bp:
Massive `config.json` overhaul!

Quick review:
* exercises are listed in difficulty order, and then by alphabetical order
* reviewed difficulty - most exercises were 1 or 2, but this isn't reflective (in my opinion) of the exercises
* add/remove topics as appropriate (and per many open issues) and alphabetise them accordingly
* added a couple more core exercises (luhn, two-fer etc.) to more evenly distribute exercise unlock
  
Fixes #61, #62, #63, #64, #65, #66, #67, #68, #69, #70, #71, #72, #73, #74, #75, #103, and #109

### DIFF
--- a/config.json
+++ b/config.json
@@ -34,7 +34,7 @@
       "uuid": "2b6d4faa-1909-422a-8eac-1682be3f1b9e"
     },
     {
-      "core": true,
+      "core": false,
       "difficulty": 1,
       "slug": "two-fer",
       "topics": [
@@ -43,6 +43,7 @@
         "strings",
         "text_formatting"
       ],
+      "unlocked_by": "null",
       "uuid": "0cfac255-6871-4588-a16b-98f7692bfdbe"
     },
     {
@@ -68,7 +69,7 @@
       "uuid": "ca5139b4-8b2f-44ea-8d83-0b8ca7674436"
     },
     {
-      "core": true,
+      "core": false,
       "difficulty": 2,
       "slug": "error-handling",
       "topics": [
@@ -77,18 +78,8 @@
         "stdout",
         "strings"
       ],
+      "unlocked_by": "null",
       "uuid": "b2b4ee35-108f-45ce-b294-c10d332e5579"
-    },
-    {
-      "core": false,
-      "difficulty": 2,
-      "slug": "gigasecond",
-      "topics": [
-        "dates",
-        "mathematics"
-      ],
-      "unlocked_by": "armstrong-numbers",
-      "uuid": "176c1a86-28b1-4848-9f87-bd6ae10db6ba"
     },
     {
       "core": true,
@@ -101,7 +92,7 @@
       "uuid": "68103f44-b442-48e6-b2b5-09001f73e926"
     },
     {
-      "core": true,
+      "core": false,
       "difficulty": 2,
       "slug": "leap",
       "topics": [
@@ -109,6 +100,7 @@
         "control_flow_conditionals",
         "input_validation"
       ],
+      "unlocked_by": "null",
       "uuid": "e24451fd-761d-4d20-81d9-e470486ec44a"
     },
     {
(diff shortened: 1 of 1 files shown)


sjwarner-bp:
I think this is pretty good - let me know what you think as soon as, and then we can re-assess where we are in terms of a launch 🎉 🎆

budmc29:
My opinion on this is that we should get at least 2-3 opinions for each exercise regarding difficulty and topics, and when we reach something that looks like an agreement we can commit to those numbers/strings.

I'll give my suggestions in a few days if this is still open.

I don't see any harm in merging everything right now, but let's see what others think.

budmc29:
Existing ratings (not including your commits):

hello-world = 1 1
gigasecond = 3 4 4
bob = 3 5
leap = 1
raindrops = 1
difference-of-squares = 1
pangram = 2
anagram = 4
hamming = 2
rna-transcription = 2
two-fer = 1
nucleotide-count = 2
word-count = 3 8
phone-number = 2
error-handling = 1
atbash-cipher = 5

sjwarner-bp:
Why do some have more than one?

budmc29:
Based on all the ratings, from other people that are on the existing opened PR.

sjwarner-bp:
Ah OK cool, makes sense to me!

budmc29:
Looks like we'll also have to wait before we agree on the core exercises as well in https://github.com/exercism/bash/issues/109, before integrating this one.

budmc29:
It might be work following the request from: https://github.com/exercism/ruby/pull/689

>If possible, leave 3 or 4 simple exercises as (core: false, unlocked_by: null), as this will provide new participants with some exercises that they can tackle even if they have not finished the first core exercise.

sjwarner-bp:
Makes sense to me. Which few exercises do we think we should leave as (core: false, unlocked_by: null)? I'll then amend this and we can keep this up to date with all our latest thinking 🙂

budmc29:
`two-fer` is good, and let's add `error-handling` for obvious reasons, and `leap-year` for some math and date manipulation.

sjwarner-bp:
Cool, I'll change this PR to reflect those in a bit

config.json
```diff
@@ -16,189 +28,206 @@
...
+        "strings"
       ],
-      "uuid": "c9d36155-10e8-4a75-a732-3c43a68910eb"
+      "uuid": "b2b4ee35-108f-45ce-b294-c10d332e5579"
     },
     {
       "core": false,
-      "difficulty": 1,
-      "slug": "acronym",
+      "difficulty": 2,
```
budmc29: We should change this to 4.
We have the following scores: 2 3 4 4.

budmc29: We should change this to 4.
We have the following scores: 2 3 4 4.

sjwarner-bp:
Cool, I'll change that with the `core: false` and `unlocked_by: null` properties soon.

We want to keep this from going stagnant! :slightly_smiling_face:

budmc29:
With those changes and if we complete #109 , this is pretty much ready to be merged.

sjwarner-bp:
I agree. I think that this is probably the largest thing stopping the track from being launched at this moment in time.

sjwarner-bp:
I've altered the core exercises as in #109 and have made the changes discussed previously here too. Think this could be ready to merge.

budmc29:
Yes, looks good. Thank you very much for your effort and patience on this one.

PR #130, opened 2018-01-04
Add the correct command to run the tests in README.md

budmc29:


budmc29:
Don't worry, you don't need to write the same message in multiple places, I am subscribed to receive notifications for this track.

They seem related, but are different. This PR updates the README.md files to show the correct `bats` command, and #131 renames the plural version of the commands and the tests names themselves.
For example this introduces `bats anagram_tests.sh` and #131 changes it to `bats anagram_test.sh` (observe the missing 's').

As for `bats whatever_test.sh` being inserted, I don't know if we can make it somehow to pick up the correct name, so it's fine as it is, we'll just have to keep an eye for that when integrating PR's with exercises.

sjwarner-bp:
> They seem related, but are different. This PR updates the README.md files to show the correct bats command, and #131 renames the plural version of the commands and the tests names themselves.

Thanks for explaining.

> As for `bats whatever_test.sh` being inserted, I don't know if we can make it somehow to pick up the correct name.

When using configlet to generate READMEs (as we should), it will just copy the contents of our insert. As such, I don't think there is a way to get the slug name (and I don't think it's too necessary, as it's probably obvious to users that `whatever_test.sh` doesn't exist).

Maybe the improvement we need to make here is to indicate `whatever` is just some placeholder - maybe `<slug_name>_test.sh`? What do you think?

budmc29:
> Maybe the improvement we need to make here is to indicate whatever is just some placeholder - maybe <slug_name>_test.sh? What do you think?

Yes, that sounds like something that would make it harder to forged to update and can be caught easier by the implementer and the reviewer.

PR #133, opened 2018-01-04
[armstrong-numbers] Update tests to check for output

budmc29:
Fixed the following with the existing armstrong-number exercise:

- Test for output instead of exit status
- Fix wrong broken test case
- Fix example implementation


budmc29:
Do you have any links for discussion? I know Ruby has a `BookKeeping` constant, but I could not find any general discussion in my searches.

sjwarner-bp:
I think often this is track specific, so there isn't a lot of discussion about a way to go about this specifically.

Coming from the Java track myself, we stick a `version` file in to each exercise, and then occasionally will run a script that checks against the problem specifications to indicate if an exercise is out of date. I think this is a really good way to do it.

If I recall correctly, F# have test suite generators that can make test files? I'm not too sure about this though.

kotp:
A bit of history for the `BookKeeping` is that it was hard to reconcile "This doesn't pass the tests" without having some indication as to if it was meant to pass the _current_ test suite or not.

There has been some desire to be able to do away with the module itself, and therefore the "version" that is handwritten by the student.  I believe this is going to be the case with Nexercism... but I have not been able to be very involved in that.  Perhaps this is something that [exercism/configlet](https://github.com/exercism/configlet) is going to handle for us?

sjwarner-bp:
Unless I am unclear, this is for feedback to a student (e.g. a student checking an old solution).

Apologies if I was unclear, but my suggestion was a `version` file (maybe it should be `.version`) as a reference to the canonical data that the `slug_name_test.sh` implements. This would be for maintainers/members of @exercism/bash, allowing us to update the test suites as and when their canonical data is changed.

budmc29:
That sound like something that might help up.
I don't think it's related to this PR since that's a new feature, and requires some planning and implementation, while this PR adds something we already agreed on.

exercises/armstrong-numbers/armstrong_numbers_test.sh
```diff
@@ -1,58 +1,66 @@
 #!/usr/bin/env bats
 
-@test 'Single digits are armstrong numbers' {
-  #skip
-  run ./armstrong_numbers.sh isarmstrong 5
+@test 'Single digits are Armstrong numbers' {
+  # skip
+  run ./armstrong_numbers.sh is_armstrong 5
```
kotp: I like the underscore to delimit the names, rather than `isArmstrong` though, we should follow bash conventions here.

If this is convention, then I am fine with it.

exercises/armstrong-numbers/example.sh
```diff
@@ -1,19 +1,25 @@
...
+    echo "true"
+    exit 0
   else
-    return 1
+    echo "false"
+    exit 1
   fi
 }
+
+"$@"
```
kotp: I won't do review on the solutions, as these are not seen by the students, and as long as they pass the tests, they are "valid" as a proof of concept.

sjwarner-bp:
> That sound like something that might help us.
> I don't think it's related to this PR since that's a new feature, and requires some planning and implementation, while this PR adds something we already agreed on, maybe open an issue for it?

Sure thing - I'll open an issue 🙂

sjwarner-bp:
Thanks for this @budmc29

exercises/armstrong-numbers/example.sh
budmc29: I'd be very happy to get a review here: http://exercism.io/submissions/8ad0a5430e594fbaa8a5c7015af43a9e, if you have the time.

kotp: Peeking.

Issue #104, opened 2017-12-26
Run exercise tests through Travis


sjwarner-bp:
As mentioned in #4, for each pull request, a run through all exercise tests would help maintainers quickly identify if the submission causes the tests to fail.

~Travis is currently used to lint submissions.~ <- incorrect! configlet lint is pulled in, but #116 talks about code linting.

sjwarner-bp:
I personally have little experience in Travis, but am more than happy to look in to this if no-one else has experience - let me know :+1:

Smarticles101:
I've got some experience with Travis. I might be able to work on this if I have time.

sjwarner-bp:
Sounds good to me @Smarticles101

budmc29:
@Smarticles101 any updates on this? If you are too busy I can give it a try this weekend.

sjwarner-bp:
Fixed via #148 (Thanks! :slightly_smiling_face: )

PR #145, opened 2018-01-08
[anagram] Remove trailing comma from test

budmc29:


budmc29:
Lesson learned from this:

    bats unexpected EOF while looking for matching `"'

sjwarner-bp:
Hopefully most of these issues will be ironed out when #104 is completed :slightly_smiling_face:

Issue #114, opened 2017-12-29
Track policies (with a view to create POLICIES.md)


sjwarner-bp:
It would be most useful to have a succinct list of the current track policies, to allow future contributors (and reviewers) to facilitate the most desirable contributions.

This would also be a great reference tool, and would be a useful part of a template for PRs.

Before this PR, I'd like a discussion to take place regarding current policies (older members of @exercism/bash may have a better idea of longer-standing policies) and policies we should strive to implement/achieve.

sjwarner-bp:
It seems like a good starting point would be "Plain Bash over POSIX compliance", as decided in #88.

sjwarner-bp:
Another could be to formally record an onboarding process / document how we will ensure that onboarding is as easy as possible.

The Java track (where I have most experience) do this by having a number of good-first-patch issues open at any one time.

sjwarner-bp:
Further, documenting the decision to come out of #87 would be beneficial to future contributors.

budmc29:
This can be added: https://github.com/exercism/bash/issues/88#issuecomment-354574633

budmc29:
Test structure can be added: #150

PR #154, opened 2018-01-29
Add Policy file

budmc29:
Fixes: #114


POLICIES.md
```diff
@@ -0,0 +1,36 @@
...
+
+  - Don't enforce POSIX compliance for our internal code, or code submitted as the solution and we allow users to use any utilities as long as the test passes on their system.
+  - Have POSIX compliance as a stretch goal
+  - For the stretch goal that could be a good start for a discussion when users submit their solutions.
+
+References: [[1](https://github.com/exercism/bash/issues/88)]
+
+### Don't include a exercise file
+
+When you submit an exercise you should only include only the test file and a README.md file.
```
sjwarner-bp: Don't we also have to supply an `example.sh`, and moving forward we've agreed for a `.meta/version` file too?

POLICIES.md
```diff
@@ -0,0 +1,36 @@
...
+
+1. your alternative perspective;
+2. your proposed changes; and
+3. how your proposal balances your own perspective with those already raised during prior discussions.
+
+## Policy Descriptions
+
+### Standard Bash over POSIX compliance
+
+  - Don't enforce POSIX compliance for our internal code, or code submitted as the solution and we allow users to use any utilities as long as the test passes on their system.
```
sjwarner-bp: > Don't enforce POSIX...
This could sound friendlier - maybe we should consider this, as the document will be read by prospective contributors?

sjwarner-bp:
Once this has been merged in, I'll create a pull request template that includes this too :slightly_smiling_face:

budmc29:
I basically copied the last comment on the issues, I agree we need a better wording for them and other missed policies.

budmc29:
We could also have a look at https://github.com/exercism/bash/blob/master/CONTRIBUTING.md, extract any policies like 'Bash style guide' and reference them in here.

POLICIES.md
budmc29: Any ideas?

sjwarner-bp:
That's a good idea - I agree

POLICIES.md
budmc29: Oh yes.

Issue #87, opened 2017-09-07
Create skeleton files for exercises


budmc29:
Do we need to create the code files for Bash? I see some languages have that implemented, and some don't.

Reference https://github.com/exercism/python/commit/de66a31bb7100ebcb2b19c0bec54184ce61ffdd8#diff-3ac5069a6087f5b553775a0720254dc1

Smarticles101:
I know on the java track we have a policy to only include skeleton files for the first like 20 exercises or so. We could potentially do something like this on the bash track, but the tests are just using testing against output of the script. I'm not sure that there's really much we could put in skeletons like method headers and such that would usually be in them on other tracks

budmc29:
I was thinking about something along these lines:
```
#!/usr/env/bin bash

function main() {
  # Add your code here
}

main
```

Smarticles101:
That seems like a good idea. We could also include variables simply getting the input passed in. Thoughts @exercism/bash?

kenden:
@Smarticles101 variables getting the input passed in, could you detail more what you're thinking about?

Also, I'd like some nice default included. [This page](http://kvz.io/blog/2013/11/21/bash-best-practices/) has a few nice ones. At least the the `set` commands?

Smarticles101:
@kenden Like `input="$1";`

budmc29:
@sjwarner-bp what is your opinion on this?

sjwarner-bp:
I think that this could be a good idea - but maybe only for the first few exercises?

On the Java track, they have chosen to put skeleton files in for all exercises under a '4' difficulty. I don't think this would work for us, because most of our exercises are below 4 (revamp and re-estimation of difficulty to come now the pending exercise PRs have been merged)!

I'd suggest maybe adding them just to core exercises. While it is useful for first-time users, part of learning to write in Bash is learning to do it from scratch :sli

budmc29:
Looks like we have reached an agreement and we are all in favor of adding a skeleton for the exercises.

> I think that this could be a good idea - but maybe only for the first few exercises?
@sjwarner-bp 

That's a very good idea, combining the both worlds. Getting the users familiar with a good script structure and they can decide to keep it and use it for the next exercises or just delete it and use whatever they like.

sjwarner-bp:
@budmc29 which approach do you think we should go with then?

I think either of 'core' exercises or 'exercises under a 4 difficulty' could work. What do you prefer?

budmc29:
[...]

PR #157, opened 2018-01-29
Add starting template

budmc29:
As discussed in: https://github.com/exercism/bash/issues/87

Opened to suggestions.


budmc29:
Added, forgot about that folder.

PR #158, opened 2018-01-29
Update grains difficulty

budmc29:
I would even argue it's a difficulty 3 exercise since you have to deal with overflow in bash.


sjwarner-bp:
LGTM - thanks @budmc29 :slightly_smiling_face:

PR #161, opened 2018-02-04
Updated roman-numerals difficulty

edwin0258:
More accurate difficulty for roman-numerals.

### DIFF
--- a/config.json
+++ b/config.json
@@ -275,7 +275,7 @@
     },
     {
       "core": false,
-      "difficulty": 6,
+      "difficulty": 4,
       "slug": "roman-numerals",
       "topics": [
         "integers",


budmc29:
I'll give the exercise a try and see how difficult it is in bash.

edwin0258:
roman-numerals was about as difficult as gigasecond or luhn for me. Definitely not the hardest problem of the available ones.

budmc29:
@kotp what do you think?

kotp:
I will say go with the 4.  The difficulty can definitely be subjective, due to the students own restrictions.  It is bash, and as such, is it a strictly "bash" solution, what limitations do you put on yourself for solving the problems.

The (strictly) bash solution does not appear to be too difficult though, relatively speaking.

budmc29:
Thank you @edwin0258

PR #162, opened 2018-02-07
scrabble-score: add to track

edwin0258:
Adding scrabble-score problem to bash track.

---

Reviewer Resources:


config.json
```diff
@@ -47,6 +47,17 @@
       ],
       "uuid": "c9d36155-10e8-4a75-a732-3c43a68910eb"
     },
+    {
+      "core": false,
```
sjwarner-bp: Seeing as this would not be a core exercise, could you suggest an `unlocked_by:` value as well? It would help us out, as this will be required for nextercism :slightly_smiling_face:

budmc29:
We also need a `.meta/version` file that contains the version of the problem from: https://github.com/exercism/problem-specifications/blob/master/exercises/scrabble-score/canonical-data.json

Ex: https://github.com/exercism/bash/tree/master/exercises/leap

budmc29:
Thank you very much.

PR #167, opened 2018-06-10
Reformat config files

kytrinyx:
This runs the configlet fmt command, which normalizes the
contents and ordering of keys in the track config and
maintainers config.

This will let us script changes to the config files without
having unnecessary noise in the diffs when submitting pull
requests.

See exercism/meta#95

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,305 +1,313 @@
 {
+  "language": "Bash",
   "active": false,
+  "blurb": "",
   "checklist_issue": 4,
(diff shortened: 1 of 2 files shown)


kytrinyx:
@sjwarner-bp It should have, but failed because of type mismatch that I totally missed. It's fixed in https://github.com/exercism/configlet/pull/138, and I'll cut a new release today and update the PR.

PR #172, opened 2018-06-29
Replace tldp.org with other resources per #171

rpdelaney:
In #171 we discussed some reasons why tldp.org is not a reliable resource for shell scripting beginners. This PR replaces tldp.org in the `RESOURCES.md` for bash with much better guides and FAQs that will put newbies on the right track.

---

Reviewer Resources:


docs/RESOURCES.md
```diff
@@ -1,3 +1,7 @@
-- [Bash man page](http://linux.die.net/man/1/bash)  
-- [The Linux Documentation Project Guides](http://tldp.org/guides.html)  
-- [Shellcheck provides shell script guidelines and help](http://www.shellcheck.net/)  
+- [Bash man page](http://linux.die.net/man/1/bash)
```
geirha: That manual at linux.die.net looks rather outdated. I'd remove it. The [GNU Bash Manual](http://gnu.org/s/bash/manual) covers everything the man-page does anyway, and the user can just run `man bash` on their own system to get the manual page corresponding to their installed version.

rpdelaney: This makes sense to me. Anybody who has bash installed has a bash manpage at their fingertips, so why link to one that might not correspond to what they have?

@budmc29 what do you think?

kotp: I will pretty often refer to the local tools such as man or a languages CLI tool to refer to documentation, so I wholeheartedly agree in referring to these.  No everyone has 24/7 access to the internet and having access regardless is a great advantage.

budmc29: @geirha, @rpdelaney the funny thing is that my Mac's bash is the 2005 version (can't check the version on my linux machine, but I never had to update bash in the past, so I assume it's the same).

I'd replace the link to the online man page with an instruction on how to access the local bash man page (I know it's sounds easy, but we need to keep in mind that there are people starting this track not knowing what a man page is).

budmc29: @kotp I'm curios what version is your local man page?

kotp: `GNU Bash 4.3`

When I ask for the version I get:

```plain
± % bash --version                                                       !10055
GNU bash, version 4.3.11(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
----- %< -----
```

rpdelaney: > I'd replace the link to the online man page with an instruction on how to access the local bash man page

Depending on how much detail you want this to have, this is going to be rather platform dependent. Do you intend to assume that everyone is using OSX? Or would just telling people "enter `man bash` in a terminal" be enough?

budmc29: You can do whatever you like, but of course we don't want to single out only an OS, so something general should be enough.

rpdelaney: 5a4c659

budmc29:
Looks good, thank you for your contribution.

Issue #168, opened 2018-06-12
Ensure Bash track is ready for v2 launch


kytrinyx:
There are a number of things we're going to want to check before the v2 site goes live. There are notes below that flesh out all the checklist items.

- [x] The track has a page on the v2 site: https://v2.exercism.io/tracks/bash
- [x] The track page has a short description under the name (not starting with `TODO`)
- [x] The "About" section is a friendly, colloquial, compelling introduction
- [x] The "About" section follows the formatting guidelines
- [x] The code example gives a good taste of the language and fits within the boundaries of the background image
- [x] There are exercises marked as `core`
- [x] Exercises have rough estimates of difficulty
- [x] Exercises have topics associated with them
- [x] The first exercise is `auto_approve: true`

## Track landing page

The v2 site has a landing page for each track, which should make people want to join it. If the track page is missing, ping `@kytrinyx` to get it added.

### Blurb
If the header of the page starts with `TODO`, then submit a pull request to https://github.com/exercism/bash/blob/master/config.json with a `blurb` key.  Remember to get [configlet](https://github.com/exercism/configlet/releases/latest) and run `configlet fmt .` from the root of the track before submitting.

### About section
If the "About" section feels a bit dry, then submit a pull request to https://github.com/exercism/bash/blob/master/docs/ABOUT.md with suggested tweaks.

#### Formatting guidelines
In order to work well with the design of the new site, we're restricting the formatting of the `ABOUT.md`. It can use:

* Bold
* Italics
* Links
* Bullet lists
* Number lists

Additionally:

* Each sentence should be on its own line
* Paragraphs should be separated by an empty line
* Explicit `<br/>` can be used to split a paragraph into lines without spacing between them, however this is discouraged.

### Code example
If the code example is too short or too wide or too long or too uninteresting, submit a pull request to https://github.com/exercism/ocaml/blob/master/docs/SNIPPET.txt with a suggested replacement.

## Exercise metadata
Where the v1 site has a long, linear list of exercises, the v2 site has organized exercises into a small set of required exercises ("core").

If you update the track config, remember to get [configlet](https://github.com/exercism/configlet/releases/latest) and run `configlet fmt .` from the root of the track before submitting.

### Topic and difficulty
Core exercises unlock optional additional exercises, which can be filtered by topic an difficulty, however that will only work if we add topics and difficulties to the exercises in the track config, which is in https://github.com/exercism/bash/blob/master/config.json

### Auto-approval
[...]

PR #174, opened 2018-07-16
Activate track

budmc29:
Activate bash track, since we have everything we need for the launch.

#168

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,8 +1,7 @@
 {
   "language": "Bash",
-  "active": false,
+  "active": true,
   "blurb": "Bash is a Unix command interpreter, for which plaintext files of commands can be made to create scripts. It's a powerful and concise way of getting stuff done!",
-  "checklist_issue": 4,
   "exercises": [
     {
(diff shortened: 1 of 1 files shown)


config.json
```diff
@@ -1,8 +1,7 @@
 {
   "language": "Bash",
-  "active": false,
+  "active": true,
   "blurb": "Bash is a Unix command interpreter, for which plaintext files of commands can be made to create scripts. It's a powerful and concise way of getting stuff done!",
-  "checklist_issue": 4,
```
budmc29: I have removed this list since I assume it's taken care of.

After searching the code base for it, I saw some references, but didn't find any tool that I can use to see the number for myself, maybe someone can point me to that if we still need that?

guygastineau:
@budmc29 Thanks for working on this PR.

config.json
kytrinyx: Yeah, we don't need this once the track is launched. (Actually, I don't think we're referencing this in the new website, so we're good to remove it either way).

PR #177, opened 2018-07-18
[Armstrong numbers] Removed unneeded function call from test, modified example.sh to match

### LINKED ISSUE #176: Problem in Armstrong numbers test
### Hi all, redoing the armstrong-numbers problem I realized two issues:
- unlike most of the other tests the `run` statement doesn't include the command bash.  This requires that the file be made executable.  This isn't a huge deal but I'm sure it confuses a lot of students and keeps them from learning from the exercise.

#### The really big problem:
- most but not all of the tests use this format ` run ./armstrong_numbers.sh is_armstrong 153`.
  The [track policies]( https://github.com/exercism/bash/blob/master/POLICIES.md) specifies that requiring a function to be called in the script from the command line should be avoided in all but  the most complex cases which require it.
- As this is the only one used in the tests I think it is in need of revision.

##### I will make a PR with the changes :smile:

guygastineau:
The stipulation for test.sh structure for this track, located [here](https://github.com/exercism/bash/blob/master/POLICIES.md), stipulates that tests will not enforce using a specific function unless it is absolutely necessary: as in triangle or atbash-cipher.
- Therefore, I have removed the enforcement of the is_armstrong function from the armstrong-numbers problem and updated the example.sh to work with the new test.

closes #176 

---

Reviewer Resources:


guygastineau:
The end of the track policies also insinuates that the test command should be...
 `run bash example.sh argument_here`
- If you all want I can update the test to meet those standards in this PR too.

budmc29:
The PR is great, thanks.

Please add a .version file as you can see in this PR: https://github.com/exercism/bash/pull/151/files#diff-1b72aad028777600dd54c19e5a2bfc8fR1, you can get the version for each exercise from here: https://github.com/exercism/problem-specifications/blob/master/exercises/armstrong-numbers/canonical-data.json

This will also be a good time to compare and make sure this exercise is on track with the problem specification document.
Currently we have a few exercises that do not follow that file since they lag behind.

guygastineau:
Thanks for pointing me toward that other PR. I will the check the problem-specifixations too. I believe this PR brings it up to date.  @budmc29 could you request the `.version` file as a change so we are tracking it?

guygastineau:
@budmc29 I added the version file and modified the test to be consistent with the other exercises.
Pending your approval I think we might be good to go here.

budmc29:
Good job, and thank you / congrats for your first contribution to bash repo.

PR #180, opened 2018-07-19
Fix bob.sh

guygastineau:
- I conformed the bob_test.sh to the .json on problem-specifications.  We have one test commented out because it behaves differently on MAC and Linux.

- We also have one extra test but I don't see how that is a problem.

---

Reviewer Resources:


guygastineau:
Sorry, I didn't realize this would try to merge my other commits too.  I think I accidentally merged the commits from my other branch into my merged master.  I'll close this and wait until #177 is merged.

Issue #150, opened 2018-01-09
Exercise test structure


budmc29:
My question here: https://github.com/exercism/problem-specifications/issues/1070#issuecomment-355414817 was to know if we need to enforce the rules of the problem-specifications in our tests.

Requiring the `property` does not look that it's something that is critical and should be enforced, my reasoning being that everyone should be free to structure their code as their wish, some people will do a big procedure, no functions, others will do multiple functions.

If you look at: https://github.com/exercism/bash/blob/master/exercises/armstrong-numbers/armstrong_numbers_test.sh, you will see that we require the user to create a function in the tests.
In contrast, for exercises like this: https://github.com/exercism/bash/blob/master/exercises/triangle/triangle_test.sh is pretty much required.

What do you think about not forcing the user to add functions unless the exercises are dependent on them, ex: `triangle`, and let the user decide how to structure their implementations?

sjwarner-bp:
This is a good question.

I think that it is for the team to enforce as we feel appropriate. I know on the Java track we deviate from the canonical data slightly as and when it suits the exercise. The biggest thing we have to consider is how tests will be run. Once we enforce methods in our tests then everyone will have to program using them.

For some exercises, this may be beneficial, whereas for others it might be excessive.

Either way, I think the most important thing will be to enforce whatever decision we come to consistently. I can see benefits to using functions across the board, but would be interested to hear what everyone else thinks.

budmc29:
Can you list the benefits you can think of?
Do you see any downside?

sjwarner-bp:
Benefits:
* Breaking code down into functions is some new syntax for beginners to learn
* Could encourage reuse of code where users may not think/know about functions
* Allows us to mirror problem-specifications more closely
* Would allow later exercises (e.g. ciphers with encrypt/decrypt functions) that require multiple functions in a single file to be more consistent with other exercises if applied across the board.

Cons:
* For some very basic exercises, there is a potential for users to view functions as unnecessary (e.g. `hello-world` probably doesn't need a function, but would require one for consistency)

budmc29:
@kotp @Smarticles101 @kenden what do you guys think?

kenden:
I don't have a strong opinion on this but
- Users might confuse functions and arguments. Ex in:
`run ./armstrong_numbers.sh is_armstrong 5`
  is `is_armstrong` a function or a argument to the script?
- It's a bit weird to always call a script with the same argument, it could be default then. 
[...]

Issue #136, opened 2018-01-06
Add test versions in order to keep exercises up to date


sjwarner-bp:
I think it would be worthwhile putting a `version` file in each exercise, indicating which version of the `problem-specifications` canonical data it is based off. This would be useful as the problem specs are updated fairly frequently with new test cases. We'd then be able to keep our exercises up to date!

Coming from the Java track myself, we stick a `version` file into each exercise, and then occasionally will run a script that checks against the problem specifications to indicate if an exercise is out of date. I think this is a really good way to do it, and would pose this as my suggestion.

To be more exact, we could use [this script](https://github.com/exercism/java/blob/master/scripts/canonical_data_check.sh) almost as is, and put a `version` file in a `.meta` folder per exercise. Each `version` check / PR could even be opened as good-first-patch issues.

This would not be for 'user' benefit as much as it would be for the @exercism/bash team.

guygastineau:
@sjwarner-bp Did #140 take care of this? Should this issue be closed?

kotp:
I believe it is only a partial step to completing this.

sjwarner-bp:
#140 was just to add a script that will allow us to check versions of the tests against the canonical data. For us to be able to use this script, we need a `.meta/version` to be added to each exercise. The easiest way to do this will be to bring all tests in an exercise up to date and then label it with that particular version number.

guygastineau:
#185 is realated to this as it attempts to streamline updating tests.
- The generator could also possibly make or overwrite the version file in .meta after the test has been generated. What do you guys think?

guygastineau:
PR #186 is the first test version update done with the generator I'm building.

I should be able to get all our exercises up to date and with `./meta/version` files this week.

guygastineau:
More than half-way to closing this issue!

PR #187, opened 2018-07-25
updated difference-of-sqaures; added version file

guygastineau:
More material from the generator.  For this test I think it is important that there be headers for the different test types, because there is not much explanation about the flags.

Moreover, I actually think there should be flags for all three types.  There is a `"property"` key in canonical-data.json, but they are in **lower camelCase**!  So I assume we want more flag-like options for the bash track.  See [here](https://github.com/exercism/problem-specifications/blob/master/exercises/difference-of-squares/canonical-data.json) for the `canonical-data.json`.

#### I did not add a flag to the test type that did not have one before.
-If we can all agree that there should be one for consistency and sanity then I can add them to this PR.
- Please let me know what you guys think about this.

- Before it was set up in a different order from the canonical-data.json and told the user that setting flags was bonus.  I think they should all have them so we can use the order from the canonical.json - which I think we should do.

---

Reviewer Resources:


guygastineau:
Perhaps we should just be using the property tag from the json in place of the flags?

cttttt:
Although I'm not a core contributor, I agree.  It would provide a great opportunity for folks to learn the green-thread for options processing in shell, `getopts`.  With the present ordering of options and arguments, `getopts` will consider the argument the end-of-options and give up.

In general, the idiomatic command line interface for UNIX programs is (or was, before long options 🤦🏽‍♂️):

`[interpreter] command [options ...] [ -- ] arguments ...`

---

Would also be neat if there were exercises that touched on more end-user friendly ways to process options.

BTW, didn't realize that these exercises were all on Github.  Will dig in and contribute recommendations should I find the opportunity in future exercises.

exercises/difference-of-squares/difference_of_squares_test.sh
```diff
@@ -1,74 +1,70 @@
 #!/usr/bin/env bats
 
-@test 'difference of squares 1' {
-    #skip
-    run bash difference_of_squares.sh 1
-
-    [ "$status" -eq 0 ]
-    [ "$output" == 0 ]
+# Version = 1.2.0
```
budmc29: This is not needed, we have the version in the `.meta/version` file.

exercises/difference-of-squares/difference_of_squares_test.sh
```diff
@@ -1,74 +1,70 @@
...
+
+# Test to Square the sum of the numbers up to the given number
```
budmc29: `Test to` does not sound good, we need to use a different wording to let the user know we are referring to a group of tests

exercises/difference-of-squares/difference_of_squares_test.sh
```diff
@@ -1,74 +1,70 @@
...
 
-@test 'difference of squares 5' {
-    skip
-    run bash difference_of_squares.sh 5
-
-    [ "$status" -eq 0 ]
-    [ "$output" == 170 ]
+@test "square of sum 5" {
+  skip
+  run bash difference_of_squares.sh 5 -S
```
budmc29: Do we need to use `-s` throughout the tests?

budmc29:
> For this test I think it is important that there be headers for the different test types, because there is not much explanation about the flags.

I agree. We need comments to group the tests since bash does not offer this functionality.
We need to make sure that we let the user know that the tests are in groups in a clear way.

> Moreover, I actually think there should be flags for all three types. There is a "property" key in canonical-data.json, but they are in lower camelCase! So I assume we want more flag-like options for the bash track.

Have a look [here](https://github.com/exercism/bash/blob/master/POLICIES.md#exercise-test-structure) and [here](https://github.com/exercism/bash/issues/150#event-1446340584) to learn more about our decision of how to follow the `property` in `problem-specifications`.

This will answer the rest of your initial questions as well.

NobbZ:
Havng the switch **after** the input feels odd.

As a user of this “tool” I'd expect the interface to be like this

```
$ ./difference.sh --sum-of-squares 5
55
$ ./difference.sh --square-of-sums 5
225
$ ./difference.sh 5
170
```

Also, I had a hard time during reading the test file, to spot that there is a difference between `s` and `S` as flags… They are really hard to distinguish on first sight.

budmc29:
@cttttt I agree with suggestion to try and help the user learn more about `getopts`, that's a great idea, and I think it will be best to implement something like that in a new exercise and keep this one (and the others) clean and focused on their main responsibility.

> Would also be neat if there were exercises that touched on more end-user friendly ways to process options.

That's a great idea, if you are interested on implementing something like that let me know, and I can give you more information.

Here's a nice recent success story: https://github.com/exercism/problem-specifications/pull/1262e

cttttt:
@budmc29 I'd be interested for sure 💯...hit me up with that info and I'll give it a shot.

Also, what was that link again? I get a 404, and can't infer the PR from the list.

NobbZ:
@cttttt just remove the trailing `e`: exercism/problem-specifications#1262

guygastineau:
@NobbZ thank you for that input.

The flags `-s` and `-S` were in The old test so I used them in the template before asking here.

I think the flag fir the difference would make sense as `-d`.  What do you all think about this?

Does anyone have ideas for replacing the existing flags?

NobbZ:
As the difference is the actually intended behaviour while sum of squares and squares of sums are just implemention details, that we are asking the student to give him some guidance, I do not think that the difference needs a flag at all.

budmc29:
Do we need any flags? If we can simplify the tests, we should do that.

NobbZ:
@budmc29 How do you want to distinguish between the three `property`s from the canonical data then?

The canonical data asks for three different functions, two helpers and one “actual”…

budmc29:
@cttttt here you go: https://github.com/exercism/problem-specifications/blob/master/CONTRIBUTING.md#implementing-a-completely-new-exercise, let me know if you need any clarification.

You also have the example given earlier (the link is fixed now), and as you can see people in there are friendly and can help you while working on your implementation.

The main point is to have a clear idea what you want to achieve, and we can work together make than idea a reality.

exercises/difference-of-squares/difference_of_squares_test.sh
guygastineau: I can just have the template pull the description. That's what is coming after test to there.

budmc29:
I would go with implementing three simple functions, similar to how the people in the Ruby track did it: https://github.com/exercism/ruby/blob/master/exercises/difference-of-squares/difference_of_squares_test.rb.

That seems like an easy to implement and clear solution.

exercises/difference-of-squares/difference_of_squares_test.sh
guygastineau: The reason why I've been putting the version in the templates is because it could allow end users interacting with mentors to discover easily wether they're looking at the same version. This way, they wont need to have the repo cloned in order to know the version.

If you still think this shouldn't be here I will remove it from this test and the other ones I have already submitted

NobbZ:
But as I understand the policie here, we write a single script that is run through a terminal, not sourced.

budmc29:
Here is a bash example that explains better what I'm referring to: https://github.com/exercism/bash/blob/master/exercises/triangle/triangle_test.sh

guygastineau:
@budmc29

To clarify,
- you are suggesting converting the camelCase properties to snake_case and enforcing them as functions in the program.
- I think that solution would be fine.
- Should there be one for the difference property?

budmc29:
> you are suggesting converting the camelCase properties to snake_case and enforcing them as functions in the program?

Yes, that's what we agreed to do in situations like this.

> Should there be one for the difference property?

Yes, because that would make more sense, see my comments above.

NobbZ:
> Here is a bash example that explains better what I'm referring to: /exercises/triangle/triangle_test.sh@master

Okay, the calling convention used there is known to me as "subcommand". If it is implemented as a function or something else is an implementation detail which I do not care for as a user.

budmc29:
> If you think that users and mentors don't need easy access to the version of the test on exercism.io then I will remove the version string.

Yes, users and mentors are not concerned with the version in the `problem-specification` that's something for us, the track maintainers to use and make sure that the exercises in this track are up to date with what's in `problem-specification`.
More info [here](https://github.com/exercism/bash/issues/136)

> Is the description without 'test to' in front of it satisfactory as a header for the different test types?

Yes, that would work.

guygastineau:
Thank you @budmc29

I'll get those changes up soon.

guygastineau:
Hey everybody,
- I changed the test to conform to our decisions ion this conversation.
- I had to rework the `example.sh` file to pass the new test structure.  I tried to keep as many original components as possible.
- Hopefully you all agree with how this is set up.  Let me know.

exercises/difference-of-squares/difference_of_squares_test.sh
```diff
@@ -1,74 +1,66 @@
...
 
-@test 'difference of squares 1' {
-    #skip
-    run bash difference_of_squares.sh 1
-
-    [ "$status" -eq 0 ]
-    [ "$output" == 0 ]
+# Square the sum of the numbers up to the given number
+@test "square of sum 1" {
+  skip
```
budmc29: Comment skip

exercises/difference-of-squares/difference_of_squares_test.sh
```diff
@@ -1,74 +1,66 @@
...
-@test 'sum of squares 1' {
-    skip
-    run bash difference_of_squares.sh 1 -s
-
-    [ "$status" -eq 0 ]
-    [ "$output" == 1 ]
+# Subtract sum of squares from square of sums
+@test "difference of squares 1" {
+  skip
+  run bash difference_of_squares.sh difference_of_squares 1
```
budmc29: It's a bit weird with the name, but what do you think about using `difference` as the name of the function instead of `difference_of_squares`?
It's still weird, but less weird than having the same word twice.

budmc29:
Besides the two comment this PR is good for merge.
My last comment is not a requirement, more of a question. I'm fine to keep that as it is due the the weird naming.

exercises/difference-of-squares/difference_of_squares_test.sh
guygastineau: Thanks, I got all excited and forgot.

exercises/difference-of-squares/difference_of_squares_test.sh
guygastineau: I felt strange about it.  I think your suggestion is good here.

Will commit in a moment.

guygastineau:
I commented the first skip and I changed the test and example.sh to use difference as the name of the last tested function instead of 'difference_of_squares'.

Thanks everybody.  Getting closer to version files for all of our exercises.

budmc29:
Thank you.

guygastineau:
Sure. I'll remove the 'version' line from the other test I PR'd yesterday.

PR #188, opened 2018-07-26
removed 'version' line from luhn test

guygastineau:
Since my last PR revealed that the version should not be represented in the test file I checked which ones have it from my templates. It appears that it was only in the luhn exercise.

This PR fixes that.

---

Reviewer Resources:


budmc29:
Thank you, I missed that.

PR #189, opened 2018-07-26
updated hamming

guygastineau:
Got hamming up to date with the [canonical-data.json](https://github.com/guygastineau/problem-specifications/blob/master/exercises/hamming/canonical-data.json).

Had to change the output for failures in the `example.sh`.

Added `.meta/version` version is 2.1.0.

---

Reviewer Resources:


guygastineau:
I just realized I changed it from `[  "$output" -eq num ]` to `[ "$output" == "num" ]`.

I'll change that really fast.

guygastineau:
Since I won't be at my computer tonight through tomorrow I went ahead and added the non-standard test from before as I think it should be included.

Thanks.

budmc29:
Yes, I agree that the non default test is a good one for our bash track.

Looks great, thanks.

PR #190, opened 2018-07-27
Fix typo in hello world README.md

MattLewin:
Line 26 of README.md reads "verions"  instead of "versions." 

Obviously, this is a major issue that will cause bash newbies to be thrown off completely.

I have now fixed this, so you can all rest easy again.

---

Reviewer Resources:


budmc29:
Thank you.

This is a major issue indeed, good catch 😄

PR #191, opened 2018-07-27
Update version number for leap

guygastineau:
I made a template for updating leap in future, but as of now it is already up to date with [exercism/problem-specifications](https://github.com/guygastineau/problem-specifications/blob/master/exercises/leap/canonical-data.json).

We have four extra exercises that all make sense, so I just left the test as is.

---

Reviewer Resources:


budmc29:
Thanks.

PR #193, opened 2018-07-28
Updated pangram; added version file "1.4.1"

guygastineau:
[problem-specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/canonical-data.json)

---

Reviewer Resources:


exercises/pangram/pangram_test.sh
```diff
@@ -1,56 +1,63 @@
 #!/usr/bin/env bats
 
+# Check if the given string is an pangram
```
budmc29: It's better to add a newline below this, to make it clean the comment if for the file and not the first test.

guygastineau:
@budmc29

Do you want this PR to wait until your PR to problem-specs goes through?

I can just take pangram straight to version 1.4.1 then.

exercises/pangram/pangram_test.sh
```diff
@@ -1,56 +1,63 @@
...
 @test "sentence empty" {
-    #skip
-    run bash pangram.sh ""
-    [ "$status" -eq 1 ]
+  #skip
+  run bash pangram.sh ""
+  [ "$status" -eq 1 ]
```
budmc29: This test is missing an output validation, meaning that you can make the test pass (which is currently the case with the example file), without making the tool usable from the command line.

We had a discussion in the past that we should not rely on the exit status, but actually return something (like "false" for example).
I don't remember if it was on a PR I made or an issue, so I can't link to it.

This applies to all the tests.

budmc29: Your call with the status, I have nothing against it being used as well as long as we're outputting something.

budmc29: Your call with the status, I don't have anything against it being used, as long as we also output something.

budmc29: Your call with the status, I don't have anything against it being used, as long as we output something.

Sorry about that, silly comment system.

guygastineau: I'll be home from work in an hour, so I'll get these changes in place.

budmc29:
Thank you.

PR #199, opened 2018-07-31
updated grains test; modified example.sh to pass new test; added vers…

### LINKED ISSUE #197: Updating Grains: Include last case?
Our grains test has all but the last test case from the [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/grains/canonical-data.json).

It is the one that wants the total count for all the grains on the board.

Should I include it in my upcoming update and change the `example.sh` to pass, or should I leave the problem as is?

guygastineau:
…ion file

[canonical-data](https://github.com/exercism/problem-specifications/blob/master/exercises/grains/canonical-data.json)

closes #197

---

Reviewer Resources:


guygastineau:
Okay, I fixed the capitalization for the last test case.

Please let me know if you all want the `-A` flag for that test or some other arg option to denote returning the value of all grains om the board.

exercises/grains/grains_test.sh
```diff
@@ -3,79 +3,77 @@
...
   run bash grains.sh 65
-
   [ "$status" -eq 1 ]
-  [ "$output" = "Error: invalid input" ]
+  [ "$output" == "Error: invalid input" ]
 }
+
+@test "Returns the total number of grains on the board" {
+  skip
+  run bash grains.sh -A
```
budmc29: The problem specification has a `total` property so let's use that instead of something arbitrary.

exercises/grains/grains_test.sh
```diff
@@ -3,79 +3,77 @@
 @test "Square 1" {
```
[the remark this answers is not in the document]
budmc29: We should also improve the exercise to be closely related to the specification (example the name of the test should be "1" not "Square 1".

exercises/grains/grains_test.sh
guygastineau: sounds good

guygastineau:
@budmc29 
Thanks for the direction there.
Everything should be kosher now.

budmc29:
Awesome, thank you.

PR #201, opened 2018-08-01
leap;word-count: Replaced test run command with 'run bash' in the test suites

### LINKED ISSUE #200: leap: Unable to run tests
In the leap exercise if I run `bats leap_test.sh`, I would always get an error:
```shell
 `[ "$status" -eq 0 ]' failed
```
regardless of the actual returned value in my script.

I tried to run and [example solution](https://github.com/exercism/bash/blob/master/exercises/leap/example.sh), and it also fails with the same error.

But when I replaced 
```shell
run ./leap.sh 2015
```
with
```shell
run bash leap.sh 2015
```
everything is fine in that test case.

Is there a problem on my end? Maybe i should run bats with a different configuration?

ZapAnton:
Closes #200

### DIFF
--- a/exercises/leap/leap_test.sh
+++ b/exercises/leap/leap_test.sh
@@ -2,7 +2,7 @@
 
 @test 'year not divisible by 4: common year' {
   #skip
-  run ./leap.sh 2015
+  run bash leap.sh 2015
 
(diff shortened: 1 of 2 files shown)


budmc29:
Thank you.

PR #207, opened 2018-08-02
Added POSIX compliance section to the README

### LINKED ISSUE #204: Usage of coreutils
Is the usage of the standard utils like `cat` `sed`, `awk`, etc. permitted to complete the exercises on this track? Or every solution should be pure Bash?

Asking, because I haven't found anything in the READMEs for this question.

ZapAnton:
This PR updates the existing READMEs and closes  #204.

When I regenerated READMEs via `configlet`, several exercises got an updated README from the latest `problem-specification`. 
For this PR I have excluded those changes, and left only my own. 

Should I update exercises with latest `problem-specification` in this or separate PR?

---

Reviewer Resources:

### DIFF
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -18,5 +18,14 @@ After the first test(s) pass, continue by commenting out or removing the `skip` 
 
 {{ . }}
 {{ end }}
+
+## POSIX compliance
+On this track full POSIX compliance is not required. Therefore feel free to use any tool you need to complete
+the exercises, like `sed`, `awk`, `date` or even `python`. As long as the solution is portable between systems
+and does not require installing third party applications, it will be accepted.
+
+However, once you have completed and submitted the exercise, take a second look at you solution.
+Maybe the same results could be achieved in pure Bash, without using any external tools?
+
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -19,13 +19,14 @@ After the first test(s) pass, continue by commenting out or removing the `skip` 
 {{ . }}
 {{ end }}
 
-## POSIX compliance
-On this track full POSIX compliance is not required. Therefore feel free to use any tool you need to complete
-the exercises, like `sed`, `awk`, `date` or even `python`. As long as the solution is portable between systems
-and does not require installing third party applications, it will be accepted.
+## External utilities
+`Bash` is a language to write scripts that works closely with various system utilities,
+like [`sed`](https://www.gnu.org/software/sed/), [`awk`](https://www.gnu.org/software/gawk/), [`date`](https://www.gnu.org/software/coreutils/manual/html_node/date-invocation.html) and even other programming languages, like [`Python`](https://www.python.org/).
+This track does not restrict the usage of these utilities, and as long as your solution is portable
+between systems and does not require installing third party applications, feel free to use them to solve the exercise.
 
-However, once you have completed and submitted the exercise, take a second look at you solution.
-Maybe the same results could be achieved in pure Bash, without using any external tools?
+For an extra challenge, if you would like to have a better understanding of the language,
+try to re-implement the solution in pure `Bash`, without using any external tools.
 
 ## Submitting Incomplete Solutions
 It's possible to submit an incomplete solution so you can see how others have completed the exercise.
--- a/exercises/armstrong-numbers/README.md
+++ b/exercises/armstrong-numbers/README.md
@@ -25,13 +25,14 @@ After the first test(s) pass, continue by commenting out or removing the `skip` 
 Wikipedia [https://en.wikipedia.org/wiki/Narcissistic_number](https://en.wikipedia.org/wiki/Narcissistic_number)
 
 
-## POSIX compliance
-On this track full POSIX compliance is not required. Therefore feel free to use any tool you need to complete
-the exercises, like `sed`, `awk`, `date` or even `python`. As long as the solution is portable between systems
-and does not require installing third party applications, it will be accepted.
+## External utilities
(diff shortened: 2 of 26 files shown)


exercises/atbash-cipher/README.md
```diff
@@ -41,5 +41,14 @@ After the first test(s) pass, continue by commenting out or removing the `skip`
 
 Wikipedia [http://en.wikipedia.org/wiki/Atbash](http://en.wikipedia.org/wiki/Atbash)
 
+
+## POSIX compliance
+On this track full POSIX compliance is not required. Therefore feel free to use any tool you need to complete
+the exercises, like `sed`, `awk`, `date` or even `python`. As long as the solution is portable between systems
+and does not require installing third party applications, it will be accepted.
+
+However, once you have completed and submitted the exercise, take a second look at you solution.
```
budmc29: It might be better to reword this to make it as a suggestion not a command?

Something along the lines of telling the user that if they want an extra challenge they can try to rewrite the exercise in pure `Bash`.

Something like "If you would like to

config/exercise_readme.go.tmpl
```diff
@@ -18,5 +18,14 @@ After the first test(s) pass, continue by commenting out or removing the `skip`
 
 {{ . }}
 {{ end }}
+
+## POSIX compliance
+On this track full POSIX compliance is not required. Therefore feel free to use any tool you need to complete
```
budmc29: We need to rewrite this into something that will be easier to understand someone that is completely new to `Bash`, and even completely new to programming.
It does not hurt to be a bit more verbose, since after reading the text for the first time they will skim it on subsequent  exercises.

exercises/armstrong-numbers/README.md
```diff
@@ -5,7 +5,7 @@ An [Armstrong number](https://en.wikipedia.org/wiki/Narcissistic_number) is a nu
 For example:
 
 - 9 is an Armstrong number, because `9 = 9^1 = 9`
-- 10 is *not* an Armstrong number, because `10 != 1^2 + 0^2 = 2`
+- 10 is *not* an Armstrong number, because `10 != 1^2 + 0^2 = 1`
```
budmc29: Let's extract commits non related to the POSIX readme changes to a different PR please.

I don't know how familiar you are with git, so I'll just say that you can use `git cherry-pick` to port the existing changes to a new branch, and `checkout` the file on the existing PR to reset it to it's original state.

Let me know if you need any help.

budmc29:
Hello @ZapAnton, thank you for your contributions, they are really appreciated 😀 .

> Should I update exercises with latest problem-specification in this or separate PR?

Yes please (addressed in the review as well).

I have requested a few changes to make the text more friendly to beginners, which is the main focus on improving the track.

exercises/armstrong-numbers/README.md
ZapAnton: Sorry, haven't noticed this change.

I think i could just `reset` the last commit, reapply `configlet generate` and just remove undesirable changes, but I would be grateful, if you showed me the general way to handle this type of problem

exercises/atbash-cipher/README.md
ZapAnton: How about:

```text
For an extra challenge, if you would like to have
a better understanding of the language, try to re-implement 
the solution in pure `Bash`, without using any external tools.
```

guygastineau: That looks good to me.

config/exercise_readme.go.tmpl
ZapAnton: How about:

## External utilities
`Bash` is a language to write scripts that works closely with
various system utilities, like `sed`, `awk`, `date`  and even 
other programming languages, like `Python`. This track does not 
restrict the usage of this utilities, and as long as your solution is portable
between systems and does not require installing third party applications,
feel free to use them to solve the exercise.

exercises/armstrong-numbers/README.md
guygastineau: I just checked [here](https://github.com/guygastineau/problem-specifications/blob/master/exercises/armstrong-numbers/description.md) and this change was updated in the description.md for armstrong-numbers in `exercism/problem-specifications.

Does that mean it should stay in this PR?

config/exercise_readme.go.tmpl
guygastineau: I think that is much clearer.  "this" should be "these" before "utilities"

exercises/atbash-cipher/README.md
budmc29: Yep, that sounds very good.

config/exercise_readme.go.tmpl
budmc29: That sounds good, can you add links for `sed`, `awk`, `date` and `Python` to proper resources (man page or website)?

exercises/armstrong-numbers/README.md
budmc29: @guygastineau It's good to add the changes to this track, but those are not related to this PR, which was created to inform the user about third party utilities.

exercises/hello-world/README.md
```diff
@@ -23,7 +23,7 @@ likely going through bash.
 
 The downside to this is that there isn't much of a development
 ecosystem around bash like there is for other languages, and there are
-multiple versions of bash that can be frustratingly incompatible. Luckily
+multiple verions of bash that can be frustratingly incompatible. Luckily
```
budmc29: Typo

exercises/armstrong-numbers/README.md
budmc29: For this case doing `git checkout HEAD _filename_` should be enough to revert the few files that were updated.
You can then run the `configlet` on a new branch if you wish.

ZapAnton:
Done, @budmc29

budmc29:
Looks good to me, what do you think @kotp ?

budmc29:
Looks good to me, I'll leave this open for a few more hours for others to have a look at it, and I'll merge it after.

PR #220, opened 2018-08-06
updated bob to version 1.2.1

guygastineau:
Bob got a minor update.
I went ahead and made the template, as I hadn't before.
I kept our no input case in the override file.

Unfortunately some lines need double quotes and some need single.
The template file would be too cluttered if I put the logic for that there, so it has to be done by hand afterwards.

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/bob/canonical-data.json)

---

Reviewer Resources:


budmc29:
Thank you.

PR #225, opened 2018-08-09
triangle: formatted test; changed exit status to 0 when "false"

guygastineau:
Since we return "true" or "false" as a string I don't think there is any reason for the program to return 1 if the answer is "false". The script did, after all, do its job and determine if the input was or wasn't a triangle.
I think returning 1 should be reserved for real errors, like not enough arguments or the arguments aren't the right type.

Therefore, I changed the exit codes for all current tests to be 0.
I changed one line in the example.sh so it could pass this change.

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/triangle/canonical-data.json)

Let me know what you think.

---

Reviewer Resources:


exercises/triangle/triangle_test.sh
```diff
@@ -1,143 +1,133 @@
...
   run bash triangle.sh equilateral 5 4 6
-
-  [ "$status" -eq 1 ]
-  [ "$output" = "false" ]
+  [ "$status" -eq 0 ]
+  [ "$output" == "false" ]
 }
 
-@test "all zero sides illegal, so the triangle is not equilateral" {
+@test "All zero sides are illegal, so the triangle is not equilateral" {
```
budmc29: We should probably update this to all lowercase, but it's fine for now.

budmc29:
Thank you.

exercises/triangle/triangle_test.sh
guygastineau: Yeah, I agree. I figured I would open a PR in `problem-specifications` for that soon.

guygastineau:
Thank you.

PR #227, opened 2018-08-14
gigasecond: changed test dexcriptions to match canonical-data.json; a…

### LINKED ISSUE #136: Add test versions in order to keep exercises up to date

guygastineau:
…dded version file

I made the test descriptions conform to the values in its [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/gigasecond/canonical-data.json).

Unfortunately, making a template for this will be difficult because our test expects a format with information that is missing from the canonical data.  So, for now there is no way to automate updating this test.

Once this is merged we will have a version file for all exercises and we can close #136 ! 

---

Reviewer Resources:


budmc29:
Thank you.

It's great that we now have a version file on all our exercises, good job 🎉

PR #229, opened 2018-08-15
Unify maintanance scripts

budmc29:
Now that all our exercises have a `.meta/version` file, I'll update our CI to check if the exercises are up to date on each PR or something along those lines.


guygastineau:
Cool. I think this is a great idea.

bin/fetch_configlet
```diff
@@ -4,26 +4,26 @@ LATEST=https://github.com/exercism/configlet/releases/latest
...
 OS=$(
 case $(uname) in
-    (Darwin*)
-        echo "mac";;
-    (Linux*)
-        echo "linux";;
-    (Windows*)
-        echo "windows";;
-    (*)
-        echo "linux";;
```
guygastineau: Thank you for standardizing style here.

sjwarner-bp:
This is an interesting idea, though I'm not sure how much I agree with it - the Java track (where I originally took the `version` idea from) doesn't do this, and I'm not sure if we need to either.

Is the idea to just add some logging to indicate how many of our test suites are up to date?

Or is the idea that if we aren't up to date, we will get a failing build? I think that if this is the case, we should re-think. I'd argue that the build isn't 'failing' at that point, we are merely slightly behind the curve, and it can be a big task (especially during, for example, Hacktoberfest) to keep in lockstep with problem-specifications. Moreover, if this is kicked off on PRs, contributors could be discouraged from helping out if they believe they have caused the build to fail (when actually a version number somewhere in a different repo changed instead).

What do you think? :slightly_smiling_face:

guygastineau:
That's a really good question that I was also wondering.  I like the idea of it simply logging the out of date ones for us.  That way it is even easier for regular contributors to notice when a update should be made.

sjwarner-bp:
@guygastineau if it is just the case of logging, I don't particularly object, but would ask the question "who will see it?"

The logs are already pretty verbose IIRC, and they aren't presented to any user unless they actively go hunting in Travis.

I'd suggest that the `canonical_data_check.sh` in the `scripts` directory checks versions for us, and probably for the same amount of effort that any user hunting in Travis would require.

Moreover, it comes with the added bonus of not needing to wait for an entire build to run.

As I say, if it is just logging, then I'm not hugely bothered either way, but think it might be a tiny bit excessive. I think as long as someone semi-regularly runs the script and opens issues for each update, the work will get done in a somewhat timely manner :slightly_smiling_face:

guygastineau:
That is a great point. I didn't realize that we couldn't make Travis post a message outside of the build logs.  I am very inexperienced with CI.

sjwarner-bp:
Maybe there is a nice way to do it - I couldn't tell you much more either :slightly_smiling_face: 

Either way, just my two cents on the topic :slightly_smiling_face:

guygastineau:
I also agree that running the `scripts/canonical_data_check.sh` regularly should get things noticed. 

I fetch upstream on prob-specs and run that script almost daily :smile:

guygastineau:
So I was just researching a bit [here](https://docs.travis-ci.com/user/notifications/), and it seems there is some cool notification configuration available for Travis. It can even send reports to slack!

One problem is that is seems to send the whole log.  

I think it would be useful/cool to have Travis build info reported to the maintainer slack channel, but I imagine the complete raw logs are a bit much.

I will look into a way to hack only the data we want from it's logs - likely not possible - and send that instead.

guygastineau:
@budmc29

Can this set the path to canonical to the master on GitHub so it is always the most recent information?

budmc29:
My initial thoughts were to run the test checks for all the exercises, but I agree that might be a bit overkill.

It will be better to just run the checks for the exercise that's changed (in case the PR deals with an exercise), I'll look for a way to do that, if that's not possible, then I'll convert this PR to a refactor only.

NobbZ:
Failing on things unrelated to a certain contribution are demotivating for pass-by or new contributors.

If at all, make it an "allowed to fail" check in the matrix, this way you'll get notice but won't scare new contributors away.

Also you should make it a daily or so check (against master only), that runs on a regular basis and then is allowed to fail to create necessary visibility.

guygastineau:
Thank you for the expertise @NobbZ

guygastineau:
It seems like one script that travis runs daily or so and is allowed to fail would be good for visibility, like @NobbZ suggested.

I think the one to run for PR's would just check that information for the changes submitted.

budmc29:
That's a good suggestion @NobbZ, I wasn't aware that Travis has a cron job: https://docs.travis-ci.com/user/cron-jobs/

I'll clean this up to make it ready for a refactor PR, and create a new PR to implement that suggestion together with what @guygastineau and @sjwarner-bp suggested as well.

Thanks.

bin/canonical_data_check
```diff
@@ -0,0 +1,112 @@
...
+num_args=$#
+
+if [ $num_args -eq 0 ]
+then
+  print_usage
+  exit 0
+fi
+
+path_to_track=
+path_to_problem_specifications=
```
guygastineau: I think it would be really nice if `path_to_track` and `path_to_problem_specifications` defaulted to `./` and `../problem-specifications` respectively.

I like keeping all of my exercism repos parallel in a directory called `exercism`.  Being able to run the script without arguments would be really nice.

We could still let users pass arguments for a custom setup, but defaulting to my suggesting would probably be a good idea.

bin/canonical_data_check
```diff
@@ -0,0 +1,112 @@
+#!/usr/bin/env bash
+
+print_usage() {
+  echo "Usage: ./canonical_data_check.sh -t path/to/track -s path/to/problem/specifications"
```
guygastineau: If we do decide to go with my suggestion for defaulting the variable on lines 21 and 22 then we could remove the getopts loop and have line 4 be `echo "Usage: ./canonical_data_check.sh <path/to/track> <path/to/problem-specifications"`

Then lines 21 and 22 could be:
```bash
path_to_track="${1:-./}"
path_to_problem_specifications="${2:-../problem-specifications}"
```

budmc29: Good idea, updated.

budmc29:
@exercism/bash can anyone approve this PR?

guygastineau:
This is looking great @budmc29.

Thank you.

budmc29:
@guygastineau there's no need, I can fix it myself.

budmc29:
Can you give me more information about the issue?

guygastineau:
Great. Yeah I just realized the default syntax doesn't work with getopts.  Simply setting the variables directly to the defaults first will let getopts override them if args are passed, but they stay correct otherwise ;)

I'll merge this when you have fixed this.

Thank you for all of your work on this.

budmc29:
Can you show me a concrete example with input and output, that will be more helpful.
Thanks

guygastineau:
Certainly.

This works if it happens before `getopts`:
```bash
path_to_track='./'
path_to_problem_specifications='../problem-sprecifications'
```
This doesn't work with getopts:

```bash
path_to_track="${1:-./}"
path_to_problem_specification="${2:-../problem_specifications}"
```
In the second example `bin/canonical-data-check.sh -s /somewhere/else`
will set the variables as follows:
`path_to_track` is `"-s"`
`path_to_problem_specifications` is `"/somewhere/else"`

guygastineau:
Does that make sense?

budmc29:
That should fix it for now.

I'll come back and clean the scripts when I'll work on adding the ci checks per exercise.

bin/canonical_data_check
```diff
@@ -0,0 +1,105 @@
...
+  exit 1
+fi
+
+if [ -z "$path_to_problem_specifications" ]
+then
+  echo "Path to problem specifications missing."
+  print_usage
+  exit 1
+fi
+
```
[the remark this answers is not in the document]
budmc29: You are correct, removed.

guygastineau:
@budmc29 
Thank you for the great work on these scripts!

PR #231, opened 2018-08-15
isogram: added new exercise

guygastineau:
[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/isogram/canonical-data.json)

Some people think this exercise is too similar to pangram.  While it is possible to make a solution wherein only one comparative operator needs changing in order to solve both, the solutions that do that are not very efficient.

They remain similar. but the optimal solution for this exercise has some striking differences from the optimal solution for pangram.

Hopefully you all like this :smile:

---

Reviewer Resources:


exercises/isogram/example.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+LOWERCASE_INPUT="${1,,}"
+
+isogram_candidate="${LOWERCASE_INPUT//[![:lower:]]}"
+
+for letter in {a..z}; do # remove aech letter only once
```
budmc29: `each`

budmc29:
All good to go after you fix the typo.

exercises/isogram/example.sh
guygastineau: Thank you.

guygastineau:
Fixed it.  Thank you for catching it ;)

guygastineau:
Thank you.

PR #238, opened 2018-08-20
diamond: new exercise

guygastineau:
Here's the proposal for the exercise diamond.

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/diamond/canonical-data.json)

I think it makes sense for the expected variables to be stored after the skips if they remain in the same file.

---

Reviewer Resources:


guygastineau:
Closes 235

exercises/diamond/.meta/version
```diff
@@ -0,0 +1 @@
+version
```
budmc29: Missing the correct value.

exercises/diamond/diamond_test.sh
```diff
@@ -0,0 +1,119 @@
...
+EOT
+)"
+  run bash diamond.sh B
+  [ "$status" -eq 0 ]
+  [ "$output" == "$expected" ]
+}
+
+@test "Smallest non-degenerate case with odd diamond side length" {
+  skip
+  expected="$(cat << EOT
```
budmc29: If we have the skip, then the command that's run and the `expected` after, it will be easier for the user to see what is the value passed to the test right away rather than scrolling.
What do you think?

budmc29: Not a blocker for the merge though.

budmc29:
Thank you.

PR #239, opened 2018-08-21
Update starting template

budmc29:
Add note for starting template since I see many students just copying and pasting as it is.

@exercism/bash


guygastineau:
I suppose I'm ready to merge this if no one else has any suggestions.

PR #240, opened 2018-08-21
Change 'bats' shebang to 'bash'

### LINKED ISSUE #236: Bash test files shebang
I'm currently working on  #234 which will add automatic checking for our files to `travis`.

One of the issues with the linter is that it's not recognizing the `#!/usr/bin/env bats` shebang.

Since we don't use the `bats` convention and we have `exercise.sh` instead of `exercise.bats`, I think we are fine to update the shebang to be `#!/usr/bin/env bash`.
That should not make any difference since we're moving away from calling the tests directly from the CLI and our users are instructed to use `bats file_name.sh`.

If no one has any objection to this, I'll create a PR later to change those.

@exercism/bash

budmc29:
closes #236

@exercism/bash

### DIFF
--- a/exercises/acronym/acronym_test.sh
+++ b/exercises/acronym/acronym_test.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bats
+#!/usr/bin/env bash
 
 @test "basic" {
   #skip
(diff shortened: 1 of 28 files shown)


guygastineau:
Nice work. I'll change this in my templates too :smile:

PR #242, opened 2018-08-23
armstrong-numbers: Replaced return status '1' with '0' for tests that expect false result

### LINKED ISSUE #241: pangram: return code of a false result
In the `pangram` exercise each test expects that the tested function returns `0` status code, regardless if the passed string is a pangram or not.

But earlier in the `armstrong-numbers` exercise if the number is not an armstrong number the tested function is expected to return `1` status code.

Should the `pangram` test suite be changed to expect `1` status code for consistency?

ZapAnton:
Applies the consistent exit codes rule discussed in the #241 (Although it seems that only `armstrong-numbers` broke them).

Closes #241 

---

Reviewer Resources:

### DIFF
--- a/exercises/armstrong-numbers/armstrong_numbers_test.sh
+++ b/exercises/armstrong-numbers/armstrong_numbers_test.sh
@@ -12,7 +12,7 @@
   skip
   run bash armstrong_numbers.sh 10
 
-  [ "$status" -eq 1 ]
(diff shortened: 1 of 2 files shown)


budmc29:
Thank you.

guygastineau:
@ZapAbton

Thank you

PR #246, opened 2018-08-30
isbn-verifier: add new exercise

guygastineau:
Thought it would be nice to get another exercise up on the track :smile:

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json)

---

Reviewer Resources:


exercises/isbn-verifier/isbn_verifier_test.sh
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env bats
```
budmc29: Please use `bash` instead of `bats`

budmc29:
Should be good to go after you update the `shebang`

guygastineau:
Oh yeah, than you

guygastineau:
Fixed the she-bang.

Since you approved it pending that fix @budmc29 I will go ahead and squash merge this.

Thank you for catching that ;)

PR #248, opened 2018-09-03
Apply styleguide to isogram

budmc29:


guygastineau:
I assume this is ready to be merged?

guygastineau:
Thank you

budmc29:
Shellcheck does complain of doing `'` anytime, so it's a bit unfortunate.
From working with it so far, we'll use it as a starting suggestion, and having to ignore some changes, since the project is not as flexible or mature as something like `rubocop` or `eslint`.

guygastineau:
That sounds reasonable.

PR #251, opened 2018-09-10
Update bob_test.sh: improve ambiguous whitespace tests

rpalo:
Currently the tests send "\n" and "\t" to test talking to Bob with only whitespace.  A student brought it up that they probably should be sending actual, expanded tabs and newlines.  Using `$'dollar single quotes'` allows Bash to expand the escape codes before sending them to `bob.sh`.  (Generally, using `echo` does this for us, but we're running 

```bash
$ bash bob.sh "\n\t\r"
```

not using `echo` so they don't get expanded.)

Bonus: this lets students use the `[[:space:]]` character class in their solutions, which is nice and clean.

---

Reviewer Resources:

### DIFF
--- a/exercises/bob/bob_test.sh
+++ b/exercises/bob/bob_test.sh
@@ -135,14 +135,14 @@
 
 @test "alternate silence" {
   skip
-  run bash bob.sh "\t\t\t\t\t\t\t\t\t\t"
+  run bash bob.sh $'\t\t\t\t\t\t\t\t\t\t'
   [ "$status" -eq 0 ]
   [ "$output" == "Fine. Be that way!" ]
 }
 
 @test "multiple line question" {
   skip
-  run bash bob.sh "\nDoes this cryogenic chamber make me look fat?\nno"
+  run bash bob.sh $'\nDoes this cryogenic chamber make me look fat?\nno'
   [ "$status" -eq 0 ]
   [ "$output" == "Whatever." ]
 }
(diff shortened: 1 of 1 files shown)


guygastineau:
As far as I know `$'` makes it use ansi c specific expansions. I believe it will still expand everything correctly with `"`.

I will have some time this afternoon to look more closely at this, but as far as I know `"` causes the expansion not echo.

I could be missing something.
I appreciate you opening this PR. I just want to do some thorough exploration to make sure the change is completely necessary.

guygastineau:
PS. I use `[: blank:]` and it gets rid of those escape sequences, so that is some other evidence that has me wanting to vet the claim.

rpalo:
Just tested it on `msys` Bash for Windows:

Showing the problem exists

I ran this on my Mac last night with the same results (although I forgot to screenshot it).  Also, I confirmed that there are no longer raw tab or newline characters in the second case:

```bash
$ [[ "$clean" == "Hellothere." ]] && echo "Match"
Match
```

Am I maybe doing something wrong?

guygastineau:
Thanks for testing platforms other than bash.

It looks like this is a positive developement for the exercise. Thank you for your work on this @rpalo.

rpalo:
You can thank @leNEKO, the mentee that brought it up, for noticing the weirdness in the first place.  :grin:

leNEKO:
:smile:

guygastineau:
Thanks @leNEKO

PR #253, opened 2018-09-17
config.json: Removed the 'stdout' tag

### LINKED ISSUE #250: The meaning of the stdout tag
Couldn't find any sort of discussion about the issue, but feel free to close it if this is a duplicate.

My question is what is the meaning behind `stdout` tag in the exercises?

Every exercise is organized to test the content of the stdout, but only a few are marked with this tag.

If this tag is used to show that some special string is expected as an output then why `nucleotide-count` or `rna-transcription` are not tagged with it?

ZapAnton:
Closes #250

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -11,7 +11,6 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "stdout",
         "strings"
       ]
     },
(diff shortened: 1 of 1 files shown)


budmc29:
Thank you.

PR #254, opened 2018-09-20
nucleotide_count: Made the exercise to be unlocked by the 'hamming' exercise

### LINKED ISSUE #247: Unlocking nucleotide count
Do you all think that nucleotide count should be unlocked by Hamming instead of Hello-world?

Personally I think that nucleotide count is easier than Hamming, but I also think it is thematically tied to Hamming.

Let me know what you think.

@exercism/bash-contributors

ZapAnton:
Closes #247

---

Reviewer Resources:


budmc29:
Thanks.

PR #258, opened 2018-10-01
add two-fer version file

sjwarner-bp:
Adds two-fer version file to the exercise.

Two-fer is currently at version 1.2.0. This file added indicates we are all up to date, and will act as a flag if the problem-specification tests are changed.

---

Reviewer Resources:


budmc29:
There was actually a file there, I merged this by mistake, but it's more of a hassle to revert than leave it as it is.
No broken production so it's fine.

PR #267, opened 2018-12-20
Hamming: update

guygastineau:
Sorry I probably should have made a branch on my fork for this PR.
I'm trying to remember git after a long semester of music.

Here is the [cannonical data](https://github.com/exercism/problem-specifications/blob/master/exercises/hamming/canonical-data.json)

Also, we'll see if it fails the last test on here.  the validate script was failing because the example uses `$0` instead of `hamming.sh`, thus it was getting `example.sh` in the response instead of `hamming.sh`.

If the example is named as it should be by students, then this doesn't make a problem.  I'm not really sure how you all want to handle this.  It makes sense and is most idiomatic to use `$0` I think.

---

Reviewer Resources:


guygastineau:
As I expected it failed the last test.  @budmc29 what do you think we should do?  I didn't change the `example.sh`, so it was already in this state.

It doesn't fail on my `bash-trach-test-generator` local repo, because the `example.sh` is named `hamming.sh` there.

Do you think I should change the `example.sh` to return 'hamming.sh' explicitly in this exercise?

guygastineau:
Also, let me know if the input for each test should be single quotes.

exercises/hamming/hamming_test.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/usr/bin/env bats
```
budmc29: Leave this as `bash` for now.

exercises/hamming/hamming_test.sh
```diff
@@ -109,5 +53,5 @@
   skip
   run bash hamming.sh
   [ "$status" -eq 1 ]
-  [ "$output" == "Usage: hamming.sh <strand1> <strand2>" ]
+  [ "$output" == "Usage: hamming.sh <string1> <string2>" ]
 }
```
budmc29: I think `disallow first strand longer` and `disallow second strand longer` are missing.

guygastineau: Those two tests are second to last, but they are present.  Working on switching to single quotes.

guygastineau:
So, example should use the name of the file and not `$0` ?

exercises/hamming/hamming_test.sh
guygastineau: got it

exercises/hamming/hamming_test.sh
guygastineau: I think I've got it now.  There is so much removed from the prob specs that it is easier to see what's happening with diff set to split.  Thanks for getting back to me.

guygastineau:
So, I think it is good to go now.

budmc29:
@guygastineau 	you can use `$0` if that show the correct filename 🤷‍♂️ .

budmc29:
Looks good, thanks.

guygastineau:
Thanks.  Yeah, I wanted to use `$0`, but that's what threw an error in the CI pipeline.  Didn't happen when the file was named hamming.sh - just when it was example obviously.

  I'll take a look at the script you made for the tests in the pipeline. Thank you for making it by the way.  I thought it replaced example with the `exercise slug` in the temporary directory.

I'll add a PR if I can.  Then I might change the hamming example back to `$0`.  Honestly I thought `$0` worked in the pipeline before.

I'm busy at school again, so my contributions will be slow, but I am in a programming course now, so I should be able to squeeze in some extra time at the computer to help this semester.

Cheers.

guygastineau:
I just reread the `validate_exercises` script again, and there is no reason why `$0` should have made that one test fail.

This is really wierd.  I will mess with it on a fork and try to figure out why.

PR #269, opened 2019-01-15
Reorder core exercises for better student and mentor experience.

### LINKED ISSUE #268: Rearrange core exercises for better student (and mentor) progression
I've been doing a lot of Bash mentoring recently, and I was thinking that Grains is probably not the best *very first* core exercise.  And then, as I was coming up with a better one, I figured I'd take a cut at rearranging them completely.

The Ruby track has a really nice progression, so I used that as my initial model, but then, considering what's easy in Ruby is not necessarily what's easy in Bash, I reworked things a little to come up with a progression a little more fitted to the Bash track.

---

Here's what I'm thinking:

## 1. Hello World

Given

## 2. Two Fer

I like this as a second exercise for much the same reasons the Ruby and Python tracks like it.  It's a nice gentle onboarding to the mentor review process, it helps mentors work out really basic issues with students totally new to Bash, and it introduces string interpolation, which is perhaps one of the most fundamental skills in Bash -- more so than other languages.

## 3. Raindrops

This is a good first introduction to conditionals.  Since conditionals can behave a bit differently in Bash (using `-eq` instead of `==`, double or single square brackets (or arithmetic parens!), and a `then` keyword, it seems like a good fit.

## 4. Hamming

This is a good introduction to controlled looping and brace-powered string manipulation, as well as a bit of error handling!  I like it because the exercise itself isn't that complicated, which allows the student ample brainspace to figure out the mechanics of looping and indexing.

## 5. Acronym

This exercise takes skills from previous challenges and makes things a little more difficult.  It also adds some splitting and replacing, which aren't super easy to do in pure Bash, but rewarding.  I could also see this exercise pushed further back, but Ruby has it in the core, and I think I agree with that decision.

## 6. Armstrong Numbers

This exercise builds on the looping/indexing skills from Hamming, but it introduces some more complex mathematics.  The fact that it doesn't immediately follow *Grains* should keep students from going straight for `bc` too, which is a bonus.

## 7. Pangram

More looping, but what you do in each loop gets a little more complex.  Possibly a good introduction to regex comparisons in Bash too.  This one seems like a midterm, reviewing the skills of 4-6 and making you work harder.

## 8. Bob

Bob is harder in Bash than it is in most languages due to boolean variables/functions working differently.  This will build on Pangram's intro to Regex pretty solidly, and it has the opportunity to introduce the function syntax: how to build them, how to call them, and how to incorporate them into `if` statements.

## 9. Scrabble Score

I just really like this exercise.  It also is a good time to introduce Associative arrays in Bash.  I could see this exercise possibly moving forward or back one or two positions without much issue.

## 10. Grains

I hate this exercise so much, but it's likely because it's currently the very first one people encounter, regardless of skill level.  Putting it back at #10 will help reduce the number of people that make it here.  That being said, it *is* an interesting challenge, introducing overflows, showing students that it's OK to reach for external commands if they haven't yet, and finally showing of the powers of `printf`.

## 11. Luhn

Remains second to last.  This is a good junior capstone-like exercise.  A lot of moving parts, helper functions probably needed, but the logic is reasonably straightforwardly spelled out.  This might be a good time to introduce a little more formal script carpentry if it hasn't been addressed yet: structured help functions, version number, argument parsing, documentation comments, etc.  I don't know.  When I went through the Bash track, I *really* enjoyed taking my time on these last two exercises, polishing them up and making them nice.

## 12. Atbash Cipher

Remains last.  A good capstone project: difficult, combining a lot of different skills, logic that can be abstracted and encapsulated, non-trivial string chunking, and multiple possible argument combinations.  Again, possibly a good opportunity for a nice, polished, production-grade script, for those that are into that kind of thing.

---

Anyways, that's what I was thinking.  I'm open to reordering, and "no we don't want to do that" is fine too.  I just thought I'd bring it up and see if anybody thought it was a good idea.  :smiley:

rpalo:
Updated config.json with a new ordering of core exercises.  Brought
some new core exercises in as well.

New order:

1. Hello World - Same as before
2. Two-Fer - Upgraded from Side to Core
3. Error Handling - Upgraded from Side to Core
4. Raindrops - Upgraded from Side to Core
5. Hamming - Previous Core, moved earlier in track
6. Acronym - Upgraded from Side to Core
7. Armstrong Numbers - Previous Core, moved later in track
8. Pangram - Same as before
9. Bob - Same as before
10. Scrabble Score - Upgraded from Side to Core
11. Grains - Previous Core, moved much later in track
12. Luhn - Same as before
13. Atbash Cypher - Same as before

Fixes #268.  See that issue for ordering rationale.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -15,20 +15,10 @@
       ]
     },
     {
-      "slug": "reverse-string",
-      "uuid": "2b6d4faa-1909-422a-8eac-1682be3f1b9e",
-      "core": false,
-      "unlocked_by": "hello-world",
-      "difficulty": 1,
-      "topics": [
-        "strings"
-      ]
-    },
-    {
       "slug": "two-fer",
       "uuid": "0cfac255-6871-4588-a16b-98f7692bfdbe",
-      "core": false,
-      "unlocked_by": "hello-world",
+      "core": true,
+      "unlocked_by": null,
       "difficulty": 1,
       "topics": [
         "control_flow_conditionals",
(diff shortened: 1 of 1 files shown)


rpalo:
Thank you! This should make mentoring a lot easier.

mikewalker:
This explains why I was seeing four exercises that were side exercises that are now waiting for mentoring in core. :) Since I have been doing the side exercises as they were unlocked they had already been submitted.

Thanks for all you do, exercism gets better all the time.

PR #274, opened 2019-04-04
Describe skips better, with anticipation

### LINKED ISSUE #273: Exercise 1 problem
Hello! So I'm a bit confused with this first exercise. It says to comment out or remove "skip" annotations prepending other tests. What other tests are there and where are these "skip" annotations? 

Also, bats throws an error for the test shell script with the following:

/* 
 ✗ Say Hi!
   (in test file hello_world_test.sh, line 6)
     `[ "$status" -eq 0 ]' failed
*/

Is that what we are supposed to fix? (iI'm using Ubuntu on WSL for a learning environment if that information is needed)

Thanks for your help and have a great day!

kotp:
There is nothing to skip in the hello world exercise, so the comment can
be confusing for someone first seeing it.

This adds the expectation for following exercises.

closes #273

---

Reviewer Resources:

### DIFF
--- a/exercises/hello-world/README.md
+++ b/exercises/hello-world/README.md
@@ -61,7 +61,7 @@ Run the tests with:
 bats hello_world_test.sh
 ```
 
-After the first test(s) pass, continue by commenting out or removing the `skip` annotations prepending other tests.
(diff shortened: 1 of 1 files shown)


guygastineau:
Thanks.

PR #275, opened 2019-04-11
Mentor rpalo suggested a null user should work

sbonds:
Since a null user should work and produce `Hello, ` as output, adding a test for it would make the implementation clearer.

I didn't find any `canonical-data.json` for the error-handling problem specification, so I added the test directly to the `error_handling_test.sh` file.

---

Reviewer Resources:

### DIFF
--- a/exercises/error-handling/error_handling_test.sh
+++ b/exercises/error-handling/error_handling_test.sh
@@ -30,3 +30,11 @@
   [ "$status" -eq 1 ]
   [ "$output" = "Usage: ./error_handling <greetee>" ]
 }
+
+@test "empty argument" {
+  skip
+  run bash error_handling.sh ""
+
+  [ "$status" -eq 0 ]
+  [ "$output" = "Hello, " ]
+}
--- a/exercises/error-handling/error_handling_test.sh
+++ b/exercises/error-handling/error_handling_test.sh
@@ -21,6 +21,7 @@
   run bash error_handling.sh Alice Bob
 
   [ "$status" -ne 0 ]
+  [ "$output" = "Usage: ./error_handling <greetee>" ]
 }
 
 @test "print usage banner with no value given" {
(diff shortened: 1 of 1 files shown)


guygastineau:
Thank you for the pull request!

I think it is worth mentioning that these change will require `${1-world}`  instead of `${1:-world}` in the ideal solution.

I am not opposed to this. It helps teach us the intricacies of Bash.

I would like to give other mentors the chance to discuss that aspect before merging.

If there are no objections after a few days, then I will go ahead and merge it. (remind me if I forget.)

guygastineau:
closes #276

guygastineau:
closes #276

kotp:
The sample solution should include the required changes as @guygastineau  mentioned before this is merged.

sbonds:
I'm a bit puzzled why we would refer to a "Hello, world!" solution in the context of a test for the error-handling module. I don't follow what change was proposed to the `error_handling_test.sh code`. Is there a larger scope I'm missing?

guygastineau:
I was pretty tired, and I initially thought this was for hello world.  I changed ally links over, but I failed to see that this didn't default to 'world' with no arguments.

The example code passes the new tests without alteration, so I will leave it be.

Since this is a reasonable addition to the tests, and the instructions for error handling are simply to implement reasonable eroor handling for each track I will go ahead and merge this.

Sorry for the confusion ;)

guygastineau:
Thank you for fixing that existing problem.

sbonds:
In addition to the incorrect arguments output check, I also made the exit status check consistent with the usage banner check. (Previously it would accept anything nonzero, but the usage check wanted it exactly 1. I changed the incorrect arguments check to also expect a return status of exactly 1.

guygastineau:
Yes, I noticed the new exit status.  Exactly what we needed there.

Thank you for your work.

PR #281, opened 2019-05-17
port: beer-song

glennj:
This is the last one I have queued up for now.

---

Reviewer Resources:


guygastineau:
Really great work!

I would prefer if the error tests at the bottom expected a non zero retval.

Also, you can enforce expected error messages for them even if they are not included in the problem specifications repo.

glennj:
The errors *do* expect a non-zero status, the operator is `-ne`.

guygastineau:
Thanks, you are right.  I guess my eyes glossed over the `-ne`.  That'll teach me to do this from a phone :laughing: 

I'll look at your latest changes and request the eyes of another maintainer before merging.

guygastineau:
Thank you @kotp 

Thank you @glennj

PR #282, opened 2019-05-18
port: food-chain

glennj:
port of food-chain exercise

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/food-chain/canonical-data.json)

---

Reviewer Resources:


guygastineau:
Nice work!  Just a reminder, it is nice to put a link to the `canonical-data.json` file in the problem-specifications repo along with ports or any other PRs pertaining to exercises.

I really like your example solution (hilariously that matters little in the end :smiley: )

This is straight forward enough, and everything is in order, so I'll go ahead and merge it.
Well, I'll look over the source one more time first.

config.json
```diff
@@ -379,6 +379,17 @@
         "text_formatting",
         "control_flow_loops"
       ]
+    },
+    {
+      "slug": "food-chain",
+      "uuid": "4b3e89e5-a1ce-4c67-b23a-b74068885318",
+      "core": false,
+      "unlocked_by": "scrabble-score",
+      "difficulty": 3,
```
guygastineau: I think the level of complexity for a good algorithm in this exercise (especially great usage of associative arrays relating to regular arrays) warrants a higher difficulty rating for our track.

It should probably be `5`, but I would agree with anything from `4 - 6`.

guygastineau:
After changing the difficulty rating I will merge this.

Note for our future selves: I think this exercise is a good candidate for core.  It could unlock beer song given their similarities.  I think this is more interesting though.

config.json
glennj: bumped it as suggested.

guygastineau:
Thank you @glennj

PR #284, opened 2019-05-21
port: resistor-color-duo

glennj:
Good morning folks, hope your weekend was well.

Two new exercises here, very closely related -- I don't really understand why the 2nd was created since it re-uses the 1st one.

resistor-color-duo [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color-duo/canonical-data.json)

resistor-color [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color/canonical-data.json)

---

Reviewer Resources:


guygastineau:
I guess I would prefer just adding resistor-color-duo to the track.  Like you said they are very similar.

glennj:
Thanks @guygastineau . I've got lots more in the pipeline.

kotp:
> I guess I would prefer just adding resistor-color-duo to the track. Like you said they are very similar.

They are very similar because they are meant to ease the progression from the basics to a fully exposed larger "product".

Kind of similar to how "all-your-base" (if that is the name of it) is a number converter from one base to another, from binary, hex, octal, which at one point were separate, but it made sense to combine them.

config.json
```diff
@@ -404,6 +404,17 @@
         "input_validation",
         "string_comparison"
       ]
+    },
+    {
+      "slug": "resistor-color-duo",
+      "uuid": "56e4993f-6e15-44e4-bf35-a031cd94aa49",
+      "core": false,
```
kotp: If/when the other resistor color exercises show up, we would not want them to show up in the "unlocked by two-fer" area.  Otherwise it won't be clear which to finish first.

guygastineau:
Thank you for you input @kotp

I guess it seemed to me that including both would mean many students would ignore one of them.  Given that the duo version wraps the code for the first one or combining the output I think it is the more interesting exercise.

Very good example with all-your-base.
I guess they deprecated it's ancestors, because they wanted to reduce the number of trivial seeming exercises.

kotp:
Right on, and fundamentally, it (the three number base exercises) ends up being generally solvable, with really nothing new taught.  I think Resistor exercises are meant to be a longer running series of learning steps.  Maud would be able to answer this more readily, as I don't really have that information in my head at the moment.

guygastineau:
Interesting. I didn't know there were more than two of them.  I will investigate the [problem-specifications](https://github.com/exercism/problem-specifications/) when I get the chance.

kotp:
`279_Deprecate_Binary_Trinary_Octal_Hexadecimal`, so there were 4.  I forgot the trinary.

glennj:
Is there a canonical tree of which exercises are unlocked by which core exercise?

guygastineau:
Nope, it is up to the tracks to decide what makes the most sense in their language.

Given that implementations can have very different optimized algorithms across languages for the same problem I think that strategy makes sense.

I am actually interested in making this exercise core.  I should get a discussion going with the mentors about this, so it is not a maintainer only decision.

PR #285, opened 2019-05-24
port: binary-search

glennj:
Next port: binary-search

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/binary-search/canonical-data.json)

---

Reviewer Resources:


config.json
```diff
@@ -415,6 +415,16 @@
...
+      "core": false,
+      "unlocked_by": "hamming",
+      "difficulty": 4,
+      "topics": [
+        "arrays",
+        "math"
+      ]
     }
-  ]
 }
```
guygastineau: I have seen this happen before, but I can't tell what is wrong.

@kotp do you have any idea why this is failing?

exercises/binary-search/example.sh
```diff
@@ -0,0 +1,25 @@
...
+    local -i i=0
+    local -i j=$(( ${#list[@]} - 1 ))
+
+    while (( i <= j )); do
+        mid=$(( (i + j) / 2 ))
+        if (( elem == list[mid] )); then
+            echo $mid
+            return
+        elif (( elem < list[mid] )); then
+            j=$(( mid - 1 ))
```
guygastineau: I don't have a problem with this for the example solution, but strictly speaking lines 15 and 17 break from the binary search algorithms.  For it to be true binary search they need to split the remaining subslice in the middle again.

guygastineau:
What was wrong with the json? I'm curious.

glennj:
I had deleted a closing bracket on the 2nd last line.

exercises/binary-search/example.sh
guygastineau: I mean that this won't keep me from merging.

I believe that this solution takes O(n/2) time as opposed to O(log(n)) time.

guygastineau:
Okay, I guess the diff confused me.

Good work on this.  I'll look over everything's ng again when I get home.

glennj:
I've finally got my git house in order: this is the pipeline of new exercises: https://github.com/glennj/bash/branches/active

exercises/binary-search/example.sh
glennj: No, O(log2(n)) -- every step discards 1/2 the elements. It will find an element in a 128-element list in maximum 8 iterations

guygastineau: Right ho. I misread your reassignments to j using mid. I thought you were just incrementing or decrementing j.

Sorry for the confusion.

guygastineau:
Thank you for your work.

PR #299, opened 2019-05-31
port: pig-latin

glennj:
igpay atinlay

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/pig-latin/canonical-data.json)

---

Reviewer Resources:


exercises/pig-latin/README.md
```diff
@@ -0,0 +1,44 @@
...
+confusing. It obeys a few simple rules (below), but when it's spoken
+quickly it's really difficult for non-children (and non-native speakers)
+to understand.
+
+- **Rule 1**: If a word begins with a vowel sound, add an "ay" sound to the end of the word. Please note that "xr" and "yt" at the beginning of a word make vowel sounds (e.g. "xray" -> "xrayay", "yttria" -> "yttriaay").
+- **Rule 2**: If a word begins with a consonant sound, move it to the end of the word and then add an "ay" sound to the end of the word. Consonant sounds can be made up of multiple consonants, a.k.a. a consonant cluster (e.g. "chair" -> "airchay").
+- **Rule 3**: If a word starts with a consonant sound followed by "qu", move it to the end of the word, and then add an "ay" sound to the end of the word (e.g. "square" -> "aresquay").
+- **Rule 4**: If a word contains a "y" after a consonant cluster or as the second letter in a two letter word it makes a vowel sound (e.g. "rhythm" -> "ythmrhay", "my" -> "ymay").
+
+There are a few more rules for edge cases, and there are regional
```
IsaacG: Unneeded line wrap here?

exercises/pig-latin/example.sh
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env bash
+
+main() {
+    local results=()
+    for word; do
+        results+=( "$(translate "$word")" )
+    done
+    echo "${results[*]}"
```
IsaacG:
```suggestion
    echo "${results[@]}"
```

glennj: I want `*` specifically, to join the array into a single string.

exercises/pig-latin/README.md
glennj: ? Are you referring to lines 12 and 13?

guygastineau: The `REAME`'s often have lines longer than 80 chars.  This annoys me, but it is outside of the scope of the bash track repo.  They are generated by `configlet`. I think `configlet` uses a file in the `problem-specifications` repo for the meat of the `README.md`

exercises/pig-latin/example.sh
guygastineau: Many sources warn about `${myarr[*]}`, but it's use is warranted here.

Furthermore, these example solutions are not really public facing, so the same level of scrutiny desired in mentoring need not apply.

exercises/pig-latin/example.sh
```diff
@@ -0,0 +1,22 @@
...
+}
+
+translate() {
+    if [[ $1 =~ ^([aeiou]|yt|xr) ]]; then
+        echo "${1}ay"
+    elif [[ $1 =~ ^(.?qu)(.*)       ]] ||
+         [[ $1 =~ ^([^aeiou]+)(y.*) ]] ||
+         [[ $1 =~ ^([^aeiou]+)(.*)  ]]
```
guygastineau: Wow, nice use of regexp capture groups!

exercises/pig-latin/example.sh
glennj: Array expansions `${ary[*]}` and `${ary[@]}` are only different and interesting when the parameter is expanded *within double quotes*. If it's not quoted, then both act the same, and have the same issues around word splitting and pathname expansion as any unquoted variable. Believe me, this is something [I've put a lot of though into](https://stackoverflow.com/questions/12314451/accessing-bash-command-line-args-vs/12316565#12316565).

exercises/pig-latin/README.md
glennj: And it's markdown, which gets rendered into HTML, which the browser renders. So I'm not concerned about the readability of the source .md file.

exercises/pig-latin/example.sh
guygastineau: Nice link. I especially appreciate the information in the answer after yours where you clarified that the elements are joined by the first character of `IFS`.

I will have to play with that today ;)

PR #303, opened 2019-06-04
port: scale-generator

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/scale-generator/exercises/run-length-encoding/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/run-length-encoding/canonical-data.json)

---

Reviewer Resources:


guygastineau:
Good work on this.  There are a couple of things I need to bring up in a PR over at problem-specifications.

1. While it makes the exercise more difficult it is important to note that all regular scales have each note name once, and not more than once. Thus the D harmonic minor has `C3` as the leading tone (not `Db`).
2. A hexatonic scale is synthetic and doesn't require every not name, but it is of the pattern 'half, augsecond, half,  augsecond, half, augsecond`.  That is to say that a hexatonic scale in `Db` would be `Db`, `Ebb`, `F`, `Gb`, `A`, `Bb`. We can also think of them as two augmented trids placed a half step apart, ie. `Db+` and `Ebb+`.  As a general rule composers stay away from double accidentals in synthetic scales as they are unnecessary. The scale being described as hexatonic is actually called the whole tone scale (specifically it is WT1 in musical set theory).
-----
Anyway, Sorry for going on for so long. My actual major at school is composition (not CS) so I feel very strongly about this.  Once I get an issue brought up over there I will reference it from this one, so you can see its progress.

I think the worse error in the specification is the misnaming of the whole tone scale as hexatonic.  I think as a learning platform (even if not for music) that it is important for us not to disseminate misinformation.

glennj:
Perhaps we'll pause this PR while your issues get addressed. I have loads more exercises I create PRs for.

I must admit no musical education, and when I encountered this exercise in another track I was quite baffled about how to go about solving it.

guygastineau:
That sounds like a plan.  I'll see what the vibe is over at prob-spec.  The whole tone scale is an easy name change, but the harmonic minor misspelling will require significantly extra code in most track example solutions I believe.

Feel free to send any more PR's you have in the pipeline.

As always, thank you very much for your work ;)

guygastineau:
https://github.com/exercism/problem-specifications/issues/1531

glennj:
remove from queue.

PR #304, opened 2019-06-07
two-fer: unit test with two args containing spaces

IsaacG:
Test that args are handled correctly, ie "$1" over "$@" and proper quoting.

---

Reviewer Resources:


guygastineau:
You need to update the `example.sh` to pass the new test.

IsaacG:
Updated `example.sh`

guygastineau:
Thank you.

PR #308, opened 2019-06-13
implement the sieve example as purely as possible to the algorithm

glennj:
I'm an unashamed tinkerer. I want to make the example solution as close as possible to the algorithm mostly so I can demonstrate the effectiveness of simple optimizations.

---

Reviewer Resources:

### DIFF
--- a/exercises/sieve/example.sh
+++ b/exercises/sieve/example.sh
@@ -5,24 +5,22 @@
 declare -i limit=$1
 
 is_prime=()
-for ((i=0; i<=limit; i++)); do
+for ((i = 2; i <= limit; i++)); do
     is_prime[i]=true
 done
-is_prime[0]=false
-is_prime[1]=false
+
+for ((p = 2; p <= limit; p++)); do
+    # mark all multiples as not prime
+    for (( m = 2 * p; m <= limit; m += p )); do
+        is_prime[m]=false
+    done
+done
 
 primes=()
-p=2
-while ((p <= limit)); do
+for ((p = 2; p <= limit; p++)); do
     if ${is_prime[p]}; then
-        # add this prime to the list
         primes+=( $p )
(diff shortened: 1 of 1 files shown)


guygastineau:
Without tinkerers we wouldn't have the world of information we have today 😀

Would have been hilarious if you had said 'unabashed'

If I remember correctly, the original algorithm done on a tablet involved making a list from 1 to the number, and then striking though all the non prime numbers. (That is obviously without optimization)

If you think you are tempted to work on an even purer solution to this I would suggest waiting till then, so we don't have so many merges about practically the same thing.

If you think this is your last update to it, then I'll gladly merge it now 😀

glennj:
there, I'm done with it now.

guygastineau:
Thanks for your work.

exercises/sieve/example.sh
```diff
@@ -5,24 +5,22 @@
...
-for ((i=0; i<=limit; i++)); do
+for ((i = 2; i <= limit; i++)); do
     is_prime[i]=true
 done
-is_prime[0]=false
-is_prime[1]=false
+
+for ((p = 2; p <= limit; p++)); do
+    # mark all multiples as not prime
+    for (( m = 2 * p; m <= limit; m += p )); do
```
NobbZ: If I read this correctly, then it will mark all multiples of `p` as non-prime whether `p` is prime or not. This won't change the final result, but it makes the algorithm not to be the sieve of erasthostenes anymore, as this sieve would only consider primes to filter out further "candidates".

IsaacG: Yeah. There ought to be a `[[ ${is_prime[p]} = true ]] || continue`

glennj: Quite right. I'll submit another PR to fix this goof.

PR #318, opened 2019-06-17
Clean the test file: the skip search&replace was too aggressive.

glennj:
Tidy the two-fer test file.

---

Reviewer Resources:


guygastineau:
Hahaha.  Good catch. Thank you.

PR #322, opened 2019-06-19
Mostly just formatting the .md files for shorter lines.

glennj:
A note about the INSTALLATION.md changes: the `$SHELL` environment variable is set by the login system, and it contains your login shell as configured in /etc/passwd. It's not about what shell is currently running. For example
```
$ /bin/bash --login
$ echo $SHELL
/usr/local/bin/fish
```
I change that section to use the `$BASH_VERSION` shell variable instead.

---

Reviewer Resources:

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -7,7 +7,7 @@ To verify you are running a bash shell, enter the command
 
     echo $BASH_VERSION
 
-This should should output some terse information about the version of
+This should output some terse information about the version of
 bash you are running. If it does not, run the command 
 
     /bin/bash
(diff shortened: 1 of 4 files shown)


guygastineau:
Thanks

docs/INSTALLATION.md
```diff
@@ -1,5 +1,23 @@
...
+Bash is usually found on any Linux, BSD, or Unix based systems. In the
+graphical interface of your choice, look for an application named
+"Terminal". Alternatively, Ctrl+Alt+F2 will take you to a shell login
+(Ctrl+Alt+F1 or F7 to return to Graphical mode). 
 
-For Windows based Operating Systems, you may need to first install [Cygwin](https://cygwin.com/).  
+To verify you are running a bash shell, enter the command 
 
-For Apple OSX, open your Applications folder, then open the Utilities folder, finally open the Terminal application.  
+    echo $BASH_VERSION
```
kotp: Does it matter if they are using this as their Shell?  To find out the version of bash, perhaps `bash --version` should be the suggestion instead?  Even if you are using a different shell, the possibility to run a bash script is still there.

kotp: I guess I am a little late to review this...

guygastineau: @kotp you bring up a good point. If you want to PR a change I would accept it.

Mac has had bash as the default for a long time, so this information would work for most users. Users who changed their shell would likely know enough this wouldn't trip them up, but I read that Mac is changing the default shell to `zsh`? So this change might actually be necessary soon.

guygastineau: https://www.theverge.com/2019/6/4/18651872/apple-macos-catalina-zsh-bash-shell-replacement-features

PR #323, opened 2019-06-19
port: palindrome-products

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/palindrome-products/exercises/palindrome-products/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/palindrome-products/canonical-data.json)

---

Reviewer Resources:


glennj:
I have a concern about the tests for this: the "find the largest palindrome from four digit factors" test takes a long time:
```
$ time bash example.sh largest 1000 9999
99000099: [9901, 9999]
       64.22 real        63.80 user         0.31 sys
```

Should I comment this test out, or mark it as a bonus in some fashion, so the builds don't have to take an extra minute or so?

guygastineau:
That is a good question.  If you want to make it bonus I would suggest moving the tests to the end and marking `bonus` in a comment.

glennj:
The main reason I brought it up is that I don't know how big of a deal it is. Does anyone care that builds will take a bit longer?

guygastineau:
It would be a little annoying, but I think I can live with it 😀

I'll give everyone else some time to respond to this before I merge it.

guygastineau:
@kotp what are your thoughts on the increase in build time for these tests?

guygastineau:
Thank you

PR #324, opened 2019-06-20
Bash shell not required

kotp:
No need to have bash as the active shell, the goal is to write scripts, not environment settings.

---

Reviewer Resources:


kotp:
> Given that this track is for `bash` and not `sh` or `posix` I think it is better to use a 5.x version of bash. Seems silly to deprive oneself of modern features just because one's proprietary OS can't legally sign the software.

I am not using OSX and that is my version.  There isn't anything trackwide that prevents the earlier version of bash to work, right?  So I think for now, a "recent" version would happen by default...  I think making changes for a minimum version is a different issue, and perhaps not yet required.

guygastineau:
Okay, I'll merge this in.

PR #327, opened 2019-06-21
port: simple-cipher

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/simple-cipher/exercises/simple-cipher/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/canonical-data.json)

---

Reviewer Resources:


glennj:
Hey, I can't reproduce the test failure. Can someone test it out?

guygastineau:
I'll check this out when I get home this afternoon.

guygastineau:
Sorry I haven't had a chance to test this out yet.

I have a band practice about an hour away this evening too, so I likely can't get to it today.

I will try to make some time tomorrow.

@kotp if you feel like seeing what is up with this, then feel free to give it a shot.

kotp:
This does fail on my system as well.  There is also trailing whitespace that should be taken care of @glennj .  These lines:

```diff
diff --git i/exercises/simple-cipher/example.sh w/exercises/simple-cipher/example.sh
index 745d955..e0ba582 100644
--- i/exercises/simple-cipher/example.sh
+++ w/exercises/simple-cipher/example.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash 
+#!/usr/bin/env bash
 
 main() {
     local key OPTIND OPTARG
@@ -24,7 +24,7 @@ main() {
 # ord: the ascii value of a character
 # $ ord "A" #=> 65
 #
-# Note the leading single quote in the last argument 
+# Note the leading single quote in the last argument
 # https://www.gnu.org/software/bash/manual/bash.html#index-printf
 #     Arguments to non-string format specifiers (ed: such as %d)
 #     are treated as C language constants, except [...] if
```

Looking to see what I can see...

glennj:
I'm not getting any test failures on my Mac nor on a CentOS VM. I could really use a clue.

kotp:
> I'm not getting any test failures on my Mac nor on a CentOS VM. I could really use a clue.
> 
> mac bash versions: 3.2.57 and 5.0.7 (homebrew)
> centos version: 4.2.46
> 
> Every test looks like this, so I don't know what Travis is seeing:
> 
> ```
> $ bash simple_cipher.sh -k ABC encode foo; echo $?
> invalid key: contains non-lc-letter
> 1
> ```

I get:

```plain
$ bash exercises/simple-cipher/example.sh -k ABC encode foo; echo $?
fpq
0
```

kotp:
It may be due to bash version?  I am using GNU bash, version 4.4.19"

glennj:
I wonder if this is a combination of
* `$BASH_ENV` environment variable -- [6.2 Bash Startup Files](https://www.gnu.org/software/bash/manual/bash.html#Bash-Startup-Files)
* `nocasematch` shell option -- [`shopt` command](https://www.gnu.org/software/bash/manual/bash.html#index-shopt)

Are either of these things in play in your/Travis's bash environment?

```
$ cat test_case_insensitive.sh
#!/usr/bin/env bash
if [[ ABC == *[^a-z]* ]]; then
    echo "non-lower case letters detected"
else
    echo "all lower case"
fi

$ cat nocasematch.sh
shopt -s nocasematch

$ bash test_case_insensitive.sh
non-lower case letters detected

$ BASH_ENV=nocasematch.sh bash test_case_insensitive.sh
all lower case
```

kotp:
For me, I don't use bash as a shell.

These are what come up though from `shopt -s`

```plain
checkwinsize   	on
cmdhist        	on
complete_fullquote	on
expand_aliases 	on
extglob        	on
extquote       	on
force_fignore  	on
histappend     	on
interactive_comments	on
progcomp       	on
promptvars     	on
sourcepath     	on
```

glennj:
I'm at a loss. When I run it with tracing, I see
```
$ bash -x simple_cipher.sh -k ABC encode foo
+ shopt -u nocasematch
+ main -k ABC encode foo
+ local key OPTIND OPTARG
+ getopts :k: opt
+ case $opt in
+ key=ABC
+ getopts :k: opt
+ shift 2
+ [[ -z ABC ]]
+ validate_key ABC
+ local key=ABC
+ [[ ABC == *[^a-z]* ]]
+ echo 'invalid key: contains non-lc-letter'
invalid key: contains non-lc-letter
+ exit 1
```
Do you see anything different?

IsaacG:
Note that `[a-z]` is not always the same thing, depending on your LOCALE. Try using `[[:lower:]]` or setting `LC_CTYPE=C`

kotp:
It would be good to put mentor notes in, covering this problem... as I am sure it will show up on student machines.  Generally we avoid doing UTF related exercises (from what I understand from conversations in Problem Specifications and other repositories.)

glennj:
Can we determine Travis's LANG and LC* locale environment variables?

guygastineau:
I got these defaults straight from their website.

```
LANG=en_US.UTF-8
LC_ALL=en_US.UTF-8
```

guygastineau:
@glennj, thank you for your work on this.

@IsaacG, thank you for finding the solution to this frustration.

guygastineau:
@kotp, as a general rule now, I use `[[:lower:]]` instead of `a-z`.  I also try to use the other named ASCII character classes whenever possible to increase readability.

Given that things can misbehave otherwise.
It might be nice to have a note about that in the readme templates instead of just having it in the mentor notes.

I think it would still be nice in mentor notes for exercises that are especially prone to this problem.

PR #328, opened 2019-06-23
update validation script for safety

glennj:
So, I called the validate_exercises script, trying to see why the build is failing.
I did not have an "exercises_temp" directory: the `cd` failed, and the script delete my local "bash" repo.

sad face

Here's a safer version, using `mktemp` to create the tempdir. If you don't want to do that, at the very least add `mkdir -p "$temp_folder"` before trying to cd to it.

* The "cleanup" function is just my style, to make the `trap` call tidier.
* Using `ln` instead of `cp` will speed things up by a few milliseconds.

---

Reviewer Resources:

### DIFF
--- a/bin/validate_exercises
+++ b/bin/validate_exercises
@@ -6,29 +6,28 @@
 set -o errexit
 set -o nounset
 
-root_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../ && pwd )"
-temp_folder_name="exercises_temp/"
-temp_folder="$root_path/$temp_folder_name"
+temp_folder=$(mktemp -d) || { echo "cannot make temp dir"; exit 1; }
 
 # Clean up after script on normal or forced exit
-trap "{ rm -rf $temp_folder; }" EXIT
+cleanup() { rm -r "$temp_folder"; }
+trap cleanup EXIT
 
-cp exercises $temp_folder -r
-cd $temp_folder
+cd "$( dirname "${BASH_SOURCE[0]}" )/.."
+cp -r ./exercises/* "$temp_folder"
 
+cd "$temp_folder"
 for exercise in *; do
-  cd $exercise
+  cd "$exercise"
   echo "Processing $exercise"
 
   # Replace "-" with "_" to follow bash conventions
-  exercise_name=$(echo "$exercise" | sed -r 's/[-]+/_/g')
-  test_file=${exercise_name}_test.sh
+  exercise_name=${exercise//-/_}
 
   # Create implementation file from example
-  cp example.sh ${exercise_name}.sh
+  ln -f example.sh "${exercise_name}.sh"
 
   # Run the tests
-  BATS_RUN_SKIPPED=true bats $test_file
+  BATS_RUN_SKIPPED=true bats "${exercise_name}_test.sh"
 
   cd ../
 done
--- a/bin/validate_exercises
+++ b/bin/validate_exercises
@@ -6,6 +6,15 @@
 set -o errexit
 set -o nounset
 
+export BATS_RUN_SKIPPED=true    # run all the tests
+
+for tool in mktemp bats; do
+  if ! type -P "$tool" >/dev/null; then
+    echo "Can't find required tool: $tool" >&2
+    exit 1
+  fi
+done
+
 temp_folder=$(mktemp -d) || { echo "cannot make temp dir"; exit 1; }
 
 # Clean up after script on normal or forced exit
@@ -13,7 +22,9 @@ cleanup() { rm -r "$temp_folder"; }
 trap cleanup EXIT
 
 cd "$( dirname "${BASH_SOURCE[0]}" )/.."
-cp -r ./exercises/* "$temp_folder"
+
+# trailing slash: only copy directories
+cp -r ./exercises/*/ "$temp_folder"
 
 cd "$temp_folder"
 for exercise in *; do
@@ -27,7 +38,7 @@ for exercise in *; do
   ln -f example.sh "${exercise_name}.sh"
 
   # Run the tests
-  BATS_RUN_SKIPPED=true bats "${exercise_name}_test.sh"
+  bats "${exercise_name}_test.sh"
 
   cd ../
 done


guygastineau:
This is really nice work. I agree that cleanup functions help make traps nicer 😀

This is much safer than the original

guygastineau:
I think this is ready to be merged in, but for the sake of redundancy Ill see if @kotp wants a chance to review.

bin/validate_exercises
```diff
@@ -6,29 +6,28 @@
 set -o errexit
 set -o nounset
 
-root_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../ && pwd )"
-temp_folder_name="exercises_temp/"
-temp_folder="$root_path/$temp_folder_name"
+temp_folder=$(mktemp -d) || { echo "cannot make temp dir"; exit 1; }
```
kotp: Since this is an executable in `/bin/` does this translate for windows systems?

guygastineau: This is a good point.  I think this script is most important for CI/Travis, but some contributors want to run it locally first.

I think a simple dependency check for `mktemp` would satisfy me. It can simply error and tell the user that mktemp is required.

Then it is up to them to figure that out if they are on windows.

bin/validate_exercises
```diff
@@ -6,29 +6,28 @@
...
+cleanup() { rm -r "$temp_folder"; }
+trap cleanup EXIT
 
-cp exercises $temp_folder -r
-cd $temp_folder
+cd "$( dirname "${BASH_SOURCE[0]}" )/.."
+cp -r ./exercises/* "$temp_folder"
 
+cd "$temp_folder"
 for exercise in *; do
```
IsaacG: You can use `for exercise in */;` to filter for directories-only

bin/validate_exercises
```diff
@@ -6,29 +6,28 @@
...
-  cd $exercise
```
IsaacG: Using a subshell here, ie `( ... )` makes cleanup (like `cd ..` and tracking whether the first `cd "$exercise" was successful) a non-issue.

bin/validate_exercises
```diff
@@ -6,29 +6,28 @@
...
-  test_file=${exercise_name}_test.sh
+  exercise_name=${exercise//-/_}
 
   # Create implementation file from example
-  cp example.sh ${exercise_name}.sh
+  ln -f example.sh "${exercise_name}.sh"
 
   # Run the tests
-  BATS_RUN_SKIPPED=true bats $test_file
+  BATS_RUN_SKIPPED=true bats "${exercise_name}_test.sh"
```
IsaacG: Maybe just export this BATS_RUN_SKIPPED=true once at the top of the script?

guygastineau: That is a good idea. We should also consider that the current implementation groups it with the bats command.  Whenever we finally get a `--noskip` option in bats we will be changing this, and such maintenance might be simpler if this is left here.

I am up for hearing both sides of this.

guygastineau: That is a good idea. We should also consider that the current implementation groups it with the bats command.  Whenever we finally get a `--noskip` option in bats we will be changing this, and such maintenance might be simpler if this is left here.

I am up for hearing both sides of this.

bin/validate_exercises
guygastineau: I would agree with this subshell idea, except I think one extra call to the `cd` built-in is going to be faster than initializing a child shell for each test.

bin/validate_exercises
```diff
@@ -6,29 +6,28 @@
 set -o errexit
 set -o nounset
 
-root_path="$( cd "$( dirname "${BASH_SOURCE[0]}" )"/../ && pwd )"
-temp_folder_name="exercises_temp/"
```
guygastineau: Please add a dep check for mktemp that reports the problem clearly and returns a non zero status 😀

bin/validate_exercises
glennj: I thought about that, but I wasn't sure how well running the tests in a subshell would interact with the set -o and trap.

IsaacG: You might need to shove the `trap cleanup` into the loop body.
The time it takes to spin up a bash shell should not be an issue unless you're trying to validate these on a 486 processor. Premature optimization and all that.

bin/validate_exercises
glennj: done

bin/validate_exercises
glennj: done

bin/validate_exercises
glennj: I added the trailing slash to the `cp` command to only copy folders.
The small problem with `for exercise in */` is that you add a trailing slash to `$exercise` which has to then be removed.

bin/validate_exercises
glennj: done

glennj:
I know I did not break this build:
```
$ bin/fetch_configlet
gzip: stdin: not in gzip format
tar: Child returned status 1
The command "bin/fetch_configlet" failed and exited with 2 during .
Your build has been stopped.
```

bin/validate_exercises
glennj: Also, we are running with `errexit` so if `cd` fails, the script aborts anyway. I don't see the benefit of running the loop body in a subshell.

bin/validate_exercises
guygastineau: Great. Thank you.

guygastineau:
I'll try to figure this out.

guygastineau:
Our local copy of fetch_cobfiglet seems to be out of date.

Of course it has been out of date for over a month without an issue, but I'd say it would be a good idea to try it with the script from exercism/configlets/scripts

guygastineau:
Of course the error suggests we're using something that expects gzip.  The extension is `.tgz`, so that should be gzip format.  However it looks like we are getting `xz` compressed files with a `.tgz` extension.

I use `.tar.xz` quite often, but I don't decompress it the way the `fetch_configlet script does.  Have you seen anything like this anywhere else @kotp ?

guygastineau:
If we can't figure this out we might have to ask up the chain, but I would like to have our fetch script up-to-date before pinging ihid about this.

kotp:
I run `bin/configlet generate .` and it seems to run fine on my end... no errors.  And a lot of updates. :)

guygastineau:
@glennj do you want to just try this PR from a new branch after I merge in #329?

I'm really quite stumped.
You obviously haven't done any changes that should cause this issue.

kotp:
Instead of a new branch, rebase master from this branch and force push, after #329  is merged in?

guygastineau:
Thanks for the suggestion @kotp

I agree; there is no empirical evidence to suggest anything should end up differently.

There is also no empirical evidence explaining why this PR failed. (Rather I can't find the evidence of the cause)

glennj:
There we go. A temporary glitch in the configlet repo?

guygastineau:
You might be right. I thought about that too.

Oh well, at least it is all working now 😀

PR #329, opened 2019-06-25
Configlet run to update exercises

kotp:
`bin/configlet generate .` run.

---

Reviewer Resources:

### DIFF
--- a/exercises/acronym/README.md
+++ b/exercises/acronym/README.md
@@ -14,7 +14,15 @@ Run the tests with:
 bats acronym_test.sh
 ```
 
-After the first test(s) pass, continue by commenting out or removing the `skip` annotations prepending other tests.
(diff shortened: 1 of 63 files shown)


guygastineau:
So this doesn't get us any closer to understanding why `bin/fetch-configlet` failed in #328 does it?

kotp:
I believe so, I gave the argument `.` for `bin/configlet` which directs it to the `config.json` file.

guygastineau:
Right, but the CO failure is happening before configlet is even fetched from the `exercism/configlet` repo.

guygastineau:
```
$ bin/fetch_configlet
gzip: stdin: not in gzip format
tar: Child returned status 1
The command "bin/fetch_configlet" failed and exited with 2 during .
```

guygastineau:
Thank you.

PR #331, opened 2019-06-25
replace [a-z] with character class in test file, and remove double ne…

glennj:
…gation for clarity

Following up with mess of simple-cipher port, here's one instance of a character range being replaced by a character class.

---

Reviewer Resources:


guygastineau:
Thank you.

PR #332, opened 2019-06-25
port: markdown

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/markdown/exercises/markdown/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/markdown/canonical-data.json)

Note that the problem specification [wants this to be a *refactoring* exercise](https://github.com/exercism/problem-specifications/blob/master/exercises/markdown/description.md). I did not have it in me to write a crappy yet working markdown parser.

---

Reviewer Resources:


guygastineau:
Nice work.

Thank you for pointing out that your example won't correctly nest `<strong>` inside `<em>`.

At some point someone should PR to prob-spec some nesting tests so the behavior will be defined in the implementations. Until that happens though your example is fine.

guygastineau:
Does everyone think we should make this a refactoring exercise?

We could just take your example and keep it functioning while making it much worse?

I don't mind taking a stab at ruining the readability and performance of your example script 😂

IsaacG:
We don't have any refactoring exercises and refactoring is a whole different skillset/problem.
On the one hand, we can remain consistent. On the other, it might be nice to expand the skills used.

glennj:
Some thoughts about how to unfactor, aka crapify:
* no functions (one giant state machine?)
* lots and lots of code duplication
* lots of variables
* stick to good coding practices: quoting, reading from the file.
* maybe use sh constructs, `[` not `[[`, don't use `+=` assignment
    * or maybe sprinkle one or two bash syntax elements in to encourage users to discover them.
* can we avoid regex? A good markdown parser probably won't use regex anyway.
* probably want to get inspiration from python code (ruby track has not implemented this one)

glennj:
I'm going to withdraw this PR and have another crack at it.

A question: what are the mechanics of putting it into git so that when the user downloads the exercise, there is a "markdown.sh" file waiting for them? Can the repo  have a "markdown.sh"? It will get trashed during testing.

kotp:
What ever you put in the folder will be delivered to the student... With an exception of, I think, `.meta` as a folder... I would have to check.  But yes, as far as I can remember, you can deliver whatever file you want to in the exercise by placing it there.

guygastineau:
@glennj so to be clear:

You are going to try your hand at making a sloppy-but-passing solution to be refactored in addition to the better example.sh?

glennj:
That's right.

PR #337, opened 2019-06-27
port: tranpose

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/transpose/exercises/transpose/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/transpose/canonical-data.json)

---

Reviewer Resources:


guygastineau:
Nice work. I like the stdin processing.

We don't need to implement this now, but it could be fun to require that it handles stdin, files, and variables all with properly formatted text.

PR #340, opened 2019-07-04
Update exercise versions to match the problem specifications

glennj:
https://github.com/exercism/problem-specifications/tree/master/exercises

Only a few of the example solutions required patching: acronym, gigasecond, hamming

The robot-simulator is still a minor version behind, but the only change is to the test descriptions so that should have been a new patch version instead.

Code used to find differences: 
```bash
for d in bash/exercises/*; do
    [[ -d $d ]] || continue
    ex=${d##*/}
    bash_ver=$(< "$d/.meta/version")
    spec_ver=$(jq -r .version problem-specifications/exercises/$ex/canonical-data.json)
    [[ $bash_ver = "$spec_ver" ]] || printf "%-24s\t%s\t%s\n" "$ex" "$bash_ver" "$spec_ver"
done 2>/dev/null
```

---

Reviewer Resources:


guygastineau:
Thank you. By the way [canonical-data-check](https://github.com/exercism/bash/blob/master/bin/canonical_data_check) can do the check for you already 😀

PR #343, opened 2019-07-13
Port/sublist

glennj:
example [README.md](https://github.com/glennj/bash/blob/port/sublist/exercises/sublist/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json)

I kinda feel like my solution is cheating, in that I join the list into a string and use string comparison. Appreciate any thoughts.

---

Reviewer Resources:


glennj:
oof, messy commits. Let me clean that up

guygastineau:
You can just rebase and open a new pr.

By the way, I think you have a nice little solution here.

Remeber, exercism is more about idiomatic code in a specific language than purity of algorithms (or types)

I think this is an ingenious way to handle it in Bash

PR #344, opened 2019-07-13
port: sublist

glennj:
example [README.md](https://github.com/glennj/bash/blob/port/sublist/exercises/sublist/README.md)  
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json)

I kinda feel like my solution is cheating, in that I join the list into a string and use string comparison. Appreciate any thoughts.

---

Reviewer Resources:


guygastineau:
Please refer to my comment about idiomatic code over algorithmic purity as the priority.

TLDR; your solution is idiomatic, so it is fine.

glennj:
One of my hesitations with this exercise is the inherent problem parsing JSON data into a list structure. I mean, suppose the input is `'[0, 1, 2, "foo, bar\nbaz", 3, 4]'` -- what's a safe way to parse that into a bash array? If we use `jq -r '.[]'` then the output is **7** lines, not 6.

I do understand the "what if your join character already appears in the input data" problem.

I still am not 100% convinced this is an applicable exercise for the bash track.

glennj:
OK, digging a little deeper into the `jq` man page reveals the `@tsv` formatter, so I can do:

```
$ parse_json_array() {
  local data=$1
  local -n _ary=$2
  local IFS=$'\t'
  read -ra _ary < <(echo "$data" | jq -r '@tsv')
}
$ declare -a list
$ echo "$1"
[0, 1, 2, "foo, bar\nbaz\tqux", 3, 4]
$ parse_json_array "$1" list
$ declare -p list
declare -a list=([0]="0" [1]="1" [2]="2" [3]="foo, bar\\nbaz\\tqux" [4]="3" [5]="4")
```

So, solutions can be written in a safer way.

guygastineau:
So, are you wanting to change your example solution, or was that point only about the applicability of this exercise for bash?

glennj:
I'm still on the fence. 

1. The current "hacky" example.sh does work with the test cases. As a newly minted mentor, I'm worried about what solutions may appear in the future. Perhaps I shouldn't be worried, but instead excited to see what others can come up with.
1. Bash itself is incapable of parsing bash easily. But we already rely on external tools for things like sorting.
2. Pragmatically, this isn't something I'd do in bash anyway. But I'd say the same for, honestly, most of the more difficult exercises.

guygastineau:
Agreed. Honestly, the point of this track is to explore bash as a language and less so as its true application (as a shell).

I think studying bash in this way greatly improved our ability to write real world scripts.

As an odd example:
I had a series of consulting jobs for the same guy, and he wanted everything done in bash.  Perl was available, but still everything needed to be bash.

The last project I worked on I had to make a lib with an implementation script that read HTML files in a directory and parsed the server info from each one into a standardized html table.

Bash is probably the last language anyone wants to use for a job like that, but experience learning bash as a language instead of just treating it like a shell made this task relatively easy and honestly pretty fun.

kotp:
I am good with the sample solution.   I am good with the exercise.  Part of learning Bash is learning to recognize the tools you can use to solve problems... and the limitations of introducing those tools.  I am on a machine that doesn't have the `jq` "command" available, and so if I were to come across a solution that used `jq` to solve the problem, I would likely accept it, but talk about those expectations, and if it is impossible or just inconvenient to solve it using Bash only...

PR #346, opened 2019-07-15
use bash double brackets in tests

glennj:
Not a high priority item, but the mix of double brackets and single brackets in the test files has been an irritant.

Besides, the `[` command is not documented to have a `==` operator. I guess bash uses a special case for that under the hood.

---

Reviewer Resources:


guygastineau:
This has bothered me for so long, but I never remember to make it happen.

Thank you @glennj

PR #354, opened 2019-07-21
We should be doing string matching for output tests, not numeric

glennj:
I was reviewing a solution where the student was outputting an empty string, but the `[[ $output -eq 0 ]]` test was passing. My recent fix to use double brackets introduced this regression:

```
$ output=""
$ [ "$output" -eq 0 ] && echo y || echo n
bash: [: : integer expression expected
n
$ [[ $output -eq 0 ]] && echo y || echo n
y
$ [[ $output == "0" ]] && echo y || echo n
n
```

This fix updates all uses of `[[ $output -eq some_number ]]`

---

Reviewer Resources:


guygastineau:
Thank you.

PR #360, opened 2019-07-28
port: resistor-color-trio

glennj:
exercise [README.md](https://github.com/glennj/bash/blob/port/resistor-color-trio/exercises/resistor-color-trio/README.md)
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/resistor-color-trio/canonical-data.json)

---

Reviewer Resources:


guygastineau:
Nice work. I guess I should actual get you to PR resistor color duo again.  It is directly references in the readme to this exercise, and my opinion/understanding of this set of exercises has changed (more in favor).

Thank you.

guygastineau:
Well, I said to PR duo, but I can't remember which one we left out.  Obviously I meant the one to which I previously said no.

glennj:
IMO all three of these exercises can be combined into one:
* resistor-color: take one color, output a number
* resistor-color-duo: take two colors, output a 2 digit number
* resistor-color-trio: take 3 colors, use the first 2 for resistor-color-duo, raise that to 10^(3rd-color), but then output with an SI-prefix.

So each new one build upon the previous ever-so incrementally. 

I submitted resistor-color and resistor-color-duo in one PR (https://github.com/exercism/bash/pull/284), and we decided to drop resistor-color.

I think we could deprecate resistor-color-duo.

guygastineau:
Yes, you are right that they build on the previous ones. That was the initial reason that I wanted to drop the simpler `resistor-color`. @kotp 's argument was that they are meant to guide implementation through these stages.

Most tracks that have them have the precursors, so I suppose that staying with convention might be nice here.

I guess, this exercise is friendlier if someone has already done duo.  Duo is still simple enough that I don't think we need to include the first one for it to be accessible, but if you PR the first one by itself I am not against merging it.

Is that an agreeable position?

glennj:
I don't think we need the first one: I too think the second one is simple enough, and it *completely* reproduces the first one.

guygastineau:
Wlee, perfect 😁

PR #362, opened 2019-07-29
new script to aid mentors reviewing submissions

glennj:
New script to assist mentors when reviewing a student's solution.
It's interactive every step of the way, but with sensible defaults if the mentor just hits Enter at every prompt.

One potentially confusing/irritating aspect is that the script launches an interactive bash shell in the download directory so the mentor can play with the solution -- If the mentor is doing a few reviews in a row, this could lead to a few nested interactive shells. The reason is that I want to leave the mentor in the solution directory and of course the script cannot alter the parent shell.

I have tested this on a few reviews, and I'm happy with it.

---

Reviewer Resources:


mentor_tools/mentor_checks
```diff
@@ -0,0 +1,133 @@
...
+
+shopt -s extglob
+
+if ! type -p exercism >/dev/null; then
+    echo "Can't find exercism in the PATH" >&2
+    exit 1
+fi
+
+main() {
+    local uuid=$1
```
kotp: Could we take advantage of the "copy" button provided on the exercise page, allowing that full command to be copied via the "click" and pasted to this command as input to it?

glennj: How about a shell `exercism` function? I've already got one to cd into the download directory -- let's extend it for mentor download:
```
exercism() {
    if [[ $1 != "download" ]]; then
        command exercism "$@"
    elif [[ $2 =~ "--uuid="([[:alnum:]]+) ]]; then
        /path/to/bash/mentor_tools/mentor_checks "${BASH_REMATCH[1]}"
    else
        local out
        readarray -t out < <(command exercism "$@")
        printf "%s\n" "${out[@]}"
        [[ -d "${out[-1]}" ]] && cd "${out[-1]}"
    fi
}
```

glennj: We could add that as a separate file that you can source in your .bashrc

guygastineau: A wrapper could be a good solution. I guess the completion function still works on a wrapper of the same name.

At the same time I still fancy the idea of making it explicit when using this tool as opposed to exercism.

Let me think about this a bit.  I'll also pull your work to my machine and try it out.

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+#    "users/" subdirectory of your exercism workspace)
+# 2. run shellcheck on the code, if you have it installed locally
+# 3. check github for the existance of mentor notes.
+# 4. show the code and ask if you want to run the tests
+# 5. launch an interactive bash shell, so you can
+#    play with  the solution
+
+shopt -s extglob
+
+if ! type -p exercism >/dev/null; then
```
IsaacG: Assuming someone has an alias or function pointing to exercism, should that not suffice? Is the `-p` needed?

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+fi
+
+main() {
+    local uuid workdir slug root
+    local uuid_re='[[:alnum:]]{32}'
+
+    if  [[ $1 =~ ^($uuid_re)$ ]] ||
+        [[ "$1 $2 $3" =~ "exercism download --uuid="($uuid_re)$ ]]
+    then
+        uuid=${BASH_REMATCH[1]}
```
glennj: OK, here we accept the full download command.

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+        echo "usage: $0 <uuid>"
+        echo "       $0 exercism download --uuid=<uuid>"
+        echo "downloads the submission, runs tests and shellcheck"
+        exit 1
+    fi
+
+    echo "=== 1 ==="
+    download "$uuid" workdir
+
+    slug=$(basename "$workdir")
```
IsaacG: Why not use `slug="${workdir%/*}"` ?

mentor_tools/mentor_checks
glennj: Just tested with a review.

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+        uuid=${BASH_REMATCH[1]}
+    else
+        echo "usage: $0 <uuid>"
+        echo "       $0 exercism download --uuid=<uuid>"
+        echo "downloads the submission, runs tests and shellcheck"
+        exit 1
+    fi
+
+    echo "=== 1 ==="
+    download "$uuid" workdir
```
IsaacG: Is there a reason you don't just pass through `"$workdir"` directly?

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+    local -a output
+
+    yn "Download solution from https://exercism.io/mentor/solutions/$uuid" || exit
+
+    readarray -t output < <( exercism download --uuid="$uuid" 2>&1 )
+    printf "%s\n" "${output[@]}"
+    dir=${output[-1]}
+
+    if [[ ! -d $dir ]]; then
+        echo "Odd, could not find the download directory in exercism output"
```
IsaacG: s/Odd/Error:/
Also print `$dir` for debugging use.

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+        exit 1
+    fi
+}
+
+run_shellcheck() {
+    local file=$1
+    if ! type -p shellcheck >/dev/null; then
```
IsaacG: Drop the `-p` ?

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+    local slug=$1
+    local notes_url curl_opts http_status
+
+    if ! type -p curl >/dev/null; then
+        echo "Can't find curl in the PATH."
+        return
+    fi
+
+    notes_url="https://raw.githubusercontent.com/exercism/website-copy/master/tracks/bash/exercises/$slug/mentoring.md"
+    curl_opts=( --silent --head --write-out '%{http_code}' )
```
IsaacG: Drop `--head` if you don't actually want it

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+    http_status=$( curl "${curl_opts[@]}" "$notes_url" | tail -1 )
+
+    case $http_status in
+        200) echo "Mentor notes exist:"; echo "$notes_url" ;;
```
IsaacG: Maybe `echo "Mentor note: $notes_url"`

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,138 @@
...
+        *)   echo "Unexpected query status from github: $http_status" ;;
+    esac
+}
+
+run_tests() {
+    local root=$1
+    cat << END_BATS
+Next step is testing:
+I will not do that  automatically.
+*******************************************
```
IsaacG: Double space. Replace with single space.

mentor_tools/mentor_checks
glennj: because the download function populates the variable. It also emits text to stdout for the user's benefit so I can't use command substitution to return the workdir value.

mentor_tools/mentor_checks
glennj: I think that a user's aliases and functions won't be available while this script is running (unless the functions are exported, which is probably unusual). 

If you have an alias or wrapper function, you probably (but not definitely) also have exercism in your PATH.

mentor_tools/mentor_checks
glennj: Six of one, half a dozen of the other.

mentor_tools/mentor_checks
glennj: Will do.

mentor_tools/mentor_checks
glennj: I don't need to actually download the notes, I only need to check if that url exists.

mentor_tools/mentor_checks
glennj: OK. I figured printing the url allows the mentor to Ctrl+click on it, and that would be better than printing all the notes: agree?

mentor_tools/mentor_checks
IsaacG: Maybe. Unless you custom installed it. Is there any reason *not* to drop `-p` and let bash check if the command would JustWork vs making bash check if there's an executable in the PATH?

mentor_tools/mentor_checks
IsaacG: Agreed. I was just suggesting printing it all on one line. A "notes exist" output line seems to have limited value to me, vs "Here's the note URL: http://..."

mentor_tools/mentor_checks
glennj: The problem is if exercism is not in the PATH, then in a non-interactive shell, `type` will not be able to find it. Can you suggest a way around this problem?

Maybe we don't need this check at all. After all, we're expecting mentors will be using it and it's reasonable to expect a bash mentor will already have it installed, or can figure out how to manage if it's not in the PATH.

mentor_tools/mentor_checks
glennj: on second thought, the exercism output has already been printed

mentor_tools/mentor_checks
kotp: If it can't find exercism, send a message early on that exercism is not in the path?  Do it once, and do it loud.

guygastineau:
I'll finally have time to play with this this afternoon.  I am very excited by your work here.

FYI, one possibility @kotp and I were discussing us the name.  I think that short memorable names are often very nice. The current name is good, and very direct.

I think 'bantor' as in 'BAsh meNTOR' is a good candidate.  It is short (ithink easy to remember too + no special characters).  It also invokes the idea of 'banter' for me.  Given that our mentoring conversations are a type of directed, focused banter I am quite partial to this idea.

As always, it is up for discussion, so everyone please feel free to share your opinions.

glennj:
That's a good name.

https://twitter.com/codinghorror/status/506010907021828096

guygastineau:
Now that I have had time to play with the tool I have had some ideas other than a name.

I think it would be nice for this tool to create a test log.  Essentially this would just dump the outcome of the bats test with a timestamp into a 'tests.log' or 'bantor.log' file in the directory with the student's solution.  This way we can keep a record of when they passed certain parts of the tests (most students are passing all tests before they submit, but this is not always the case, and sometimes a newer iteration breaks tests).

Also, I frequently turn the user solution download dir into a git repo, so I can mess with their solution.  This way I can try out weird suggestions (and know that they work), while working with a student.  I love the subshell we're dropped into by this script, and it facilitates this kind of mentoring very well.  I think we should populate the `BATS_RUN_SKIPPED` variable in this subshell, so that the tests can be run again without having to set that ourselves.  This would be wonderful.

As I write this I realize my mind is coming up with such grandiose ideas (including having this script automatically create a `.gitignore` or autocreating a mentor branch where we can play with things).  Such additions are outside of the scope of this PR.  I think I will be satisfied with the ideas from above being implemented.

Given that some of these wider ideas fall more under the category of `bantor` and not so much in the stricter category of `mentor_checks` I am fine with this being named either for right now.

Once this is merged we can begin working together on a more generally useful `bantor` tool that incorporates this work and other useful features.

As alway @glennj thank you for all of your hard work!

guygastineau:
Last note.  You did clever work with those default answers to the questions that `mentor_checks` asks.  It would be even nicer if the defaults were handled like they are in `fdisk` or something.  That is, it says what the default is, and the default just requires pressing the return key, but overriding it also only requires entering the value you want.

I forgot to hit backspace 2 times to override the last default while test-driving this tool ;)

kotp:
Regarding the name:

> Changing the name now does two things... One it identifies the scope in which the tool is meant to work, and it puts it in a state where once the CLI is done no other work needs to be done as to the renaming of the executable.  The elongating of the name is not a problem currently, because we can document "Set up an alias for this to bantor if you would like an easier tab-completed name."

My proposal for the name is to follow the pattern `exercism-whatever`.

guygastineau:
So, I should have made it clear that you don't absolutely need to implement the logging for this to get merged.

Setting the `BATS_RUN_SKIPPED` really would be nice, and is simple.

Making the selection behavior more convenient would be a big plus, but we can do that in the future if you don't want to do it now 😁

glennj:
> So, I should have made it clear that you don't absolutely need to implement the logging for this to get merged.

I will get to that as a "next feature" item. I do find it inconvenient that past mentoring sessions disappear until the student adds an iteration or a comment.

> Setting the `BATS_RUN_SKIPPED` really would be nice, and is simple.

I don't think that's something this script has to do. If you want that in your environment, you can put it in your environment so that it will be in the spawned shell. My .bashrc contains this so I don't have to provide any arguments:

    alias bats='BATS_RUN_SKIPPED=true bats *_test.sh'


> Making the selection behavior more convenient would be a big plus, but we can do that in the future if you don't want to do it now 😁

I don't quite understand what you want here. In my last commit, I tweaked it so that the questions I expect are "y" responses, I prepopulate the  entry field, and for the "do you want to run the tests" I leave the field blank. Just hitting enter is the same as a "no". I also found it annoying to have to backspace to enter "y"

Are you suggesting something more like `read -n 1` where you just want to tap "y" or "n" without hitting enter?

guygastineau:
To your last question - I guess I grabbed your work before the commit where you changed that.  The behavior sounds like what I'm suggesting.  I'll verify today.

Hmm, well I don't leave `BATS_RUN_SKIPPED` turned on in my shell, b cause I don't always want (although I donprobably want it more often than not).

I suppose if I want this script to do it I can just modify my local version ;)

glennj:
Perhaps we could do something like this:
```bash
exec bash -i --rcfile <( . ~/.bashrc; [[ -r ~/.bantorrc ]] && . ~/.bantorrc )
```

guygastineau:
That is a good idea. I was already thinking about suggesting or implementing a sourced rc for future versions of this project.

To clarify, if we source a `bantorrc` at the beginning of the file it's variables and functions should be present in the subshell, no?  `bash -i` doesn't wipe it does it?

I'm just asking, because if the environment persists, then we don't have to source it on that line.  This would be useful for a no prompts feature.  We don't need to implement that now (I'm thinking of the future).

I am always already going to be in the web UI when I get a student's exercise id, so I normally have already read their solution and feel comfortable executing their code on my machine.  Anyway, I'm just trying to give reasons for the rc to be sourced at the head of the program.  Let me know what you think.

mentor_tools/mentor_checks
```diff
@@ -0,0 +1,149 @@
...
+    shift $((OPTIND-1))
+    prompt=${1:-"Yes or no"}
+    read -p "$prompt [$yn]? " -e -i "$def" -r ans
+    [[ ${ans#+([[:space:]])} == Y* ]]
+}
+
+# bantor config: a place to put variables and functions can be used
+# in the environment of the spawned interactive shell (such as
+# BATS_RUN_SKIPPED) - don't forget to `export`
+for rc in "$XDG_CONFIG_HOME/bantorrc" ~/.config/bantorrc ~/.bantorrc; do
```
kotp: Since the tooling is "Exercism" should we use the "Config dir:" that comes from `exercism configure`

glennj: done

guygastineau:
Nice. This is great work!

I think we should create an installation file for this at some point, but it is better for us to merge this with some instructions for installation and use.  We can make the install script later.

Is everybody feeling good about this.
I think I am ready to merge.

IsaacG:
> To clarify, if we source a `bantorrc` at the beginning of the file it's variables and functions should be present in the subshell, no? `bash -i` doesn't wipe it does it?

Only if exported.

guygastineau:
So I just realized that our rc file is `bantorrc`, but the script is still named `mentor_checks`.  Would you all prefer changing it now, so that this isn't confusing to people who weren't part of this conversation?

Also, what is that `.swp` file?  I didn't even realize it was there.

IsaacG:
A `.swp` is likely an artifact of `vim` and not adding the `.sw*` to the `.gitignore` :)

guygastineau:
I figured it was along hose lines. I suppose I would prefer that it not get committed to master 😂

glennj:
oops. I'll remove the swp file

script name: let's vote now. I prefer `bantor`

guygastineau:
I suppose I must support that which I coined.

IsaacG:
On the odd chance that anyone is waiting for me to chime in, I abstain from commenting on the name.

guygastineau:
@kotp will it break your heart if the exercism prefix is left off?

It can be revisited later.

mentor_tools/mentor_checks
kotp: Thank you!

kotp:
Let's get this in...

The reason for the `exercism-bantor` name is not in place at this time anyway, and so it does not hold a great weight.  And it hasn't been discussed with @kytrinyx or @iHiD either...

guygastineau:
@glennj that is a go.

I'll merge this once you remove the .swp file and rename to `bantor` 😀

glennj:
Ready to go.

PR #363, opened 2019-07-30
new test added to canonical data, example required update

glennj:
problem spec:

* PR https://github.com/exercism/problem-specifications/pull/1557
* [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/word-count/canonical-data.json)

---

Reviewer Resources:


guygastineau:
Thank you.

This has me thinking we should probably put the slug with : followed by 'update' or something. I had no idea what this was for at first.

I will do some work on our contributing guide and send it up as a WIP PR. Then we can all conference on it.  It is quite old, and I think we should make it our own 😁

PR #364, opened 2019-08-01
bantor: add logging

glennj:
Added logging: it's not specific log messages, but (most) of the session is dumped into the log file.

I also moved some functions around which contributes to the bulk of the diff.

---

Reviewer Resources:

### DIFF
--- a/mentor_tools/bantor
+++ b/mentor_tools/bantor
@@ -10,8 +10,12 @@
 # 5. launch an interactive bash shell, so you can
 #    play with  the solution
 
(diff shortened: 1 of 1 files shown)


guygastineau:
Nice

kotp:
Was looking forward to seeing the two commits... unfortunate that it was only one.

glennj:
What else are you hoping for @kotp ?

kotp:
Looked like there should have been two commits... one for the logging, another for the rearrangement.

PR #370, opened 2019-08-04
stub: should be executable

guygastineau:
I realized once I copied the  stub to all exercises, that they weren't executable.  I think we fixed our tests, so that none of them require an executable script, but that might not be the case.

I remember a lot of frustration the first time I came across a test requiring an executable solution script. BATS doesn't give a good error message for this.

---

Reviewer Resources:


guygastineau:
@kotp It won't show any changes here, but it does say it is executable now.

I just wanted someone else to see it before this trivial change is merged :smile:

guygastineau:
@glennj @IsaacG 

A simple confirmation from either of you would also suffice.

guygastineau:
Thank you.

kotp:
> @kotp It won't show any changes here, but it does say it is executable now.

Not true, it shows:

`100644 → 100755` as expected.

guygastineau:
Ah, I see.

PR #372, opened 2019-08-05
remove mention of stub from exercise_stub

guygastineau:
Attempting to avoid confusion for students regarding terminology in the stub file template `exercise-stub`.

---

Reviewer Resources:

### DIFF
--- a/_template/exercise_stub
+++ b/_template/exercise_stub
@@ -21,4 +21,4 @@
 #   # call main with all of the positional arguments
 #   main "$@"
 #
-# *** REMOVE THESE COMMENTS BEFORE SUBMITTING YOUR SOLUTION ***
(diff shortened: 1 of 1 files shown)


_template/exercise_stub
```diff
@@ -21,4 +21,4 @@
 #   # call main with all of the positional arguments
 #   main "$@"
 #
-# *** REMOVE THIS STUB BEFORE SUBMITTING YOUR SOLUTION FOR MENTORING ***
+# *** REMOVE THESE COMMENTS BEFORE SUBMITTING YOUR SOLUTION ***
```
kotp: `PLEASE` in capitals as well.  It won't hurt anything, but would sure be nice if they were gone.

guygastineau: Okay.

PR #373, opened 2019-08-06
WIP: Convert to Lib tests

### LINKED ISSUE #366: Switch Tests to Library Tests
@kotp and I have discussed before the possibility of running our tests against functions from sourced libs instead of testing and executable script.

With the newandate for stubs I think this is a good time to discuss whether or not we would prefer switching to this model.

-----

#### Pros:
- Allow us to standardize the API of exercises.
- Allow easier static code analysis for the auto-mentor project (I've been meaning to work on that for us).
- Encourage students to modularize code into functions (perhaps leading to better bashers)
- Might make it easier for newcomers.
#### Cons:
- Robs autonomy from the students.
- More difficult to debug. Less savvy students might have trouble realizing how to write an interim implementation script that uses their lib, or they may not realize that they can source it into their interactive environment to play with their work. (That should be in docs or the readme's)
- it could make things slightly more complicated for some students.

-----
Well, this is a big question/decision.

I welcome viewpoints from any and all of our maintainers, mentors, and students 😀

guygastineau:
This is not intended for merging at all, but rather it is intended to share ideas about how to convert our tests to library testing.  That being said, if commits only ever involve 1 exercise, then we can easily cherry pick conversions where we agree, so they can be merged in their own PR.

For now, I just did `two-fer` initial PoC and `triangle` since it is a bit more relevant.

@glennj I played with the `::` you used for name spacing in `forth`, but I could not get it to work for `triangle`. It kept returning an error status.  Anyway, I will keep investigating that, but any investigation you feel like doing would be appreciated.

I noticed that `nounset` absolutely may not be used in the implementation files for lib testing.  `source`ing the file into bats, then pollutes the bats environment.  Since bats relies on variables being unset sometimes it will not work.  We can probably use something like `bash -c "source triangle; equilateral 2 2 2"` to keep from polluting the bats environment.  We should make sure we discuss this ;)

closes #366

---

Reviewer Resources:

### DIFF
--- a/exercises/triangle/example.sh
+++ b/exercises/triangle/example.sh
@@ -1,18 +1,13 @@
 #!/usr/bin/env bash
 
 set -o errexit   # Stop script on command error
-set -o nounset   # Error out if accessing undefined variable name
+#set -o nounset   # Error out if accessing undefined variable name
 set -o pipefail  # Error out if any step in a pipe errors out
 
-if [[ $# -ne 4 ]]; then
+usage () {
   echo "Usage: $0 [equilateral | isosceles | scalene] <s1> <s2> <s3>"
   exit 2 # Improper inputs
-fi
-
-triangle_type=$1
-s1=$2
-s2=$3
-s3=$4
+}
 
 assert() {
   # Takes a numerical inequality in a string.
@@ -22,14 +17,7 @@ assert() {
 }
 
 output() {
-  exit_code=$1
-
-  if [ $exit_code -eq 0 ]; then
-    echo "true"
-  else
-    echo "false"
-  fi
-
+  [[ $1 -eq 0 ]] && echo 'true' || echo 'false'
   exit 0
 }
 
@@ -43,34 +31,28 @@ valid_triangle() {
   return 0
 }
 
-if ! valid_triangle $s1 $s2 $s3; then
-  # Sides do not meet triangle inequality requirement
-  # Given a <= b <= c and a, b, c != 0, a + b >= c
-  output 1
-fi
+validate_args () {
+  [[ $# -eq 3 ]] || usage
+  valid_triangle "$@"
+}
 
 equilateral() {
+  validate_args "$@" || output 1
   assert "$1 == $2" && assert "$1 == $3"
 
   output $?
 }
 
 isosceles() {
+  validate_args "$@" || output 1
   assert "$1 == $2" || assert "$1 == $3" || assert "$2 == $3"
 
   output $?
 }
 
 scalene() {
+  validate_args "$@" || output 1
   assert "$1 != $2" && assert "$1 != $3" && assert "$2 != $3"
 
   output $?
 }
-
-# Bash Ternary Operator:
-# (boolean value/calculation) && action if true || action if false
-# Works because of boolean shortcutting.
-# If boolean clause is false, it doesn't evaluate the other side of &&
-# If first thing is false, it evaluates the item after ||,
-# but if first two are true, doesn't bother evaluating last part.
-$triangle_type $s1 $s2 $s3 && exit 0 || exit 1
--- a/exercises/triangle/triangle_test.sh
+++ b/exercises/triangle/triangle_test.sh
@@ -1,31 +1,34 @@
 #!/usr/bin/env bash
 
+# get user code
+source triangle.sh
+
 # Test returns true if the triangle is equilateral
 
 @test "all sides are equal, equilateral" {
   #[[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 2 2 2
+  run equilateral 2 2 2
   [[ $status -eq 0 ]]
   [[ $output == "true" ]]
 }
 
 @test "any side is unequal" {
   [[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 2 3 2
+  run equilateral 2 3 2
   [[ $status -eq 0 ]]
   [[ $output == "false" ]]
 }
 
 @test "no sides are equal, equilateral" {
   [[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 5 4 6
+  run equilateral 5 4 6
   [[ $status -eq 0 ]]
   [[ $output == "false" ]]
 }
 
 @test "all zero sides is not a triangle" {
   [[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 0 0 0
+  run equilateral 0 0 0
   [[ $status -eq 0 ]]
   [[ $output == "false" ]]
 }
@@ -34,7 +37,7 @@
 
 @test "sides may be floats, equilateral" {
   [[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 0.5 0.5 0.5
+  run equilateral 0.5 0.5 0.5
   [[ $status -eq 0 ]]
   [[ $output == "true" ]]
 }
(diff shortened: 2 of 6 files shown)


guygastineau:
As I said over in #366 I wont force push to this, so feel free to clone and we can all work together on it.  

I think local work on any given exercise should be rebased before pushing back up here to make cherry picking easier.

@kotp @glennj @IsaacG Let me know if you guys would prefer having this branch belong to our exercism repo outright.  It might be nice if we could all just have our own version of this branch through our personal clones.  Anyway, just let me know ;)

guygastineau:
> travis_fold:start:before_script
[0Ktravis_time:start:1ae0802b
[0K$ bin/fetch_configlet

gzip: stdin: unexpected end of file
tar: Unexpected EOF in archive
tar: Unexpected EOF in archive
travis_time:end:1ae0802b:start=1565059734245936845,finish=1565060043499524003,duration=309253587158
[0K[31;1mThe command "bin/fetch_configlet" failed and exited with 2 during .[0m

Your build has been stopped.


I have seen this happen before. Was it here? I think it is happening, because the files are `.xz`, but tar wants to treat them like `gzip` format due to an extension.

exercises/triangle/example.sh
```diff
@@ -22,14 +17,7 @@ assert() {
...
 output() {
-  exit_code=$1
-
-  if [ $exit_code -eq 0 ]; then
-    echo "true"
-  else
-    echo "false"
-  fi
-
+  [[ $1 -eq 0 ]] && echo 'true' || echo 'false'
```
IsaacG: (( $1 == 0 )) && ...

exercises/triangle/triangle_test.sh
```diff
@@ -1,31 +1,34 @@
...
 
+# get user code
+source triangle.sh
+
 # Test returns true if the triangle is equilateral
 
 @test "all sides are equal, equilateral" {
   #[[ $BATS_RUN_SKIPPED == true  ]] || skip
-  run bash triangle.sh equilateral 2 2 2
+  run equilateral 2 2 2
```
IsaacG: I'm a fan of exporting a single `main` or `triangle` function that takes a triangle-type as a first arg.
Thoughts?

guygastineau: Given that it is a library I think main here is not right, but I could be convinced that a managing function has its place for namespacing.

This also depends on how our other endeavors into namespacing fair.

If you have some ideas for this exercise please feel free to make a PR to this branch on my fork.

If you would prefer I am interested in making a new branch on the base repo for getting through this ordeal.

I want to know other contributor opinions first though.

guygastineau: Or you can make your own PR for triangle.

exercises/triangle/example.sh
guygastineau: You are very right about that.

At least I upgraded it from `[]` 😂

guygastineau:
I just remembered that if these are bash libraries (they are not posix compliant) then they should have `.bash` as their suffix.

kotp:
The invalid and wontfix labels are added just to remind not to merge.

guygastineau:
Thank @kotp those labels are a good reminder.

Do you have an opinion on whether or not this project should be switched to an architecture wherein it is a direct branch of `exercism/bash`?

exercises/triangle/triangle_test.sh
glennj: I think this comment thread is getting to the heart of how we want to manage namespacing in our bash libs.
1. do we want any? i.e. let `source` just dump library functionality into the "global" scope
2. use a toplevel funciton with library functions accessed as subcommands, as suggested here
3. use a function prefix, like I did [in the forth example](https://github.com/exercism/bash/blob/master/exercises/forth/example.sh)

kotp:
I think it is fine where it is, until it isn't.  We can clone it down and work with it locally, as it is.

exercises/triangle/triangle_test.sh
guygastineau: I think the world of bash library modules is all cowboys.  We should see if Gnu org has any specification.

I learned that functions may have `.` in their names too.

The only problem I have is that my syntax highlighters get confused.

guygastineau: Either of you developed any stronger opinions about how to namespace?

IsaacG: We can't prevent the global namespace from being polluted. Most shell scripts don't bother with namespaces. The few that do seem to like the space::func like in glennj's example.

These exercises don't tend to require sufficient commands that a namespace is needed and a namespace doesn't avoid other non-name-spaced functions from colliding with the global namespace. Weak opinion to not bother with name spacing.

guygastineau: So, I did some more tests.
Double colon (`::`) looks nice, but I have also read that it is not guaranteed to work with older versions of bash.  I am of the opinion that there is absolutely no reason to be using bash older than v4.x, but historically we have not wanted to make `Mac` users upgrade bash against their will.

I am not against setting v4.2 as a requirement for the track.  Hell, I am on v5.0.7, but I would not make this requirement all on my own.

If we decide against enforcing a minimum Bash version, then I see no other way to namespace the functions than by using an module access function.

In order for this not to confuse students it should be present in the stub. We would need to decide if we write the whole thing, so they don't have to worry about it or not.

Then we are left with the decision about how to name the parts it calls.  Good naming convention would then make the former API functions internal and they should be preceded by a `_`.  I think this will get ugly very quickly, and it is not really exposing students to a common pattern in the shell world.  Furthermore,  it doesn't do much to help with possible collisions in the world.

While I think that `::` is ideal (even if a bit non-standard) I am feeling less comfortable with the interface function pattern.  I honestly think that normally lib sourcing is the way to go here.

If it makes us feel any better. Golang's tests are in the same packagespace as the code, so all of the functions are called without namespacing.

Also, our namespacing ideas were never going to solve issue of set options colliisions with `bats`' internal requirements.  To do that we would need to execute the run commands in a sub-shell.  This might work, but it needs to be tested.

guygastineau: @IssacG I guess we were typing at the same time :rofl: 

It seems we both came to the same conclusion (although I am very verbose as usual).

guygastineau: Looking around at other languages like Python and Ruby it is very evident tha no one else is worrying about this (but in their languages those functions will be namespaced once they are imported somewhere).  That just isn't the way of life here :smile:

IsaacG: Note bash 3 (or, at least 3.2.57) supports space::func just fine. While we may not want to enforce bash 4+, I think it's fair to expect bash 3 or newer.

guygastineau: Thanks for checking.

glennj:
FYI I have not noticed any problems with function names containing `::` in bash 3.2 on the mac

guygastineau:
So, is it possible to get new count of our positions regarding namespaces?
The only thing that still concerns me about using `::` is that it bars the solution from being posix compatible if a student wishes to do so, but that is not a big deal in my opinion.

Still my vote would be for no namespaces, since students are unlikely to encounter that so much in wild (although it does exist).

guygastineau:
Tally for namespacing:

| Contributors      | Yay | Nay |
| -------------------- | ---- | ----- | 
| @guygastineau |  :x:   | :ballot_box_with_check:     |
| @glennj             |      |      | 
| @IsaacG           |      |      |
| @kotp                |      |      |

glennj:
> So, is it possible to get new count of our positions regarding namespaces?
> The only thing that still concerns me about using `::` is that it bars the solution from being posix compatible if a student wishes to do so, but that is not a big deal in my opinion.

Since this is a bash-specific track and bash is not POSIX compatible, IMO don't care.

> Still my vote would be for no namespaces, since students are unlikely to encounter that so much in wild (although it does exist).

This is like `import library` vs `from library import *` in python. My vote is yes, but I'll abide by majority.

IsaacG:
I'll vote for yes namespace.

guygastineau:
@IsaacG  I assume you mean with `::` from our earlier conversations?

guygastineau:
@glennj I meant using a function that takes subcommands to act as the entry point for the library.

IsaacG:
Yes, I'm in favor of using the `::` for function namespaces.

glennj:
ref #376

kotp:
I feel like this is the Bash track, and so by definition not `sh` and therefore not required to be POSIX compliant.  `sh` is a specification, and Bash is a superset of `sh`.  I do agree that we should be doing the `.bash` extension, instead.  And there should be a clear explanation for that in the introductory material handed to students.

It may be interesting to have a `sh` track as well, where strict compliance is expected, but that isn't this track.   In recent times, I have noticed:

```
file $(which sh)
/bin/sh: symbolic link to dash
```

This seems to happen on Debian (and at least some derivatives).  OpenBSD seems to use something a little more Korn-y.

So the practice of "portability" using `!#/bin/sh` is not an issue here, but could also be mentioned.  

Instead, as this is the Bash track, we should use `#!/usr/bin/env bash` or at least specifically `#!/bin/bash` and let the program itself fail if Bash isn't available.  Again, the focus being "This is the Bash track".

Just some thoughts as it regards my "vote" in [Tally for Namespacing](https://github.com/exercism/bash/pull/373#issuecomment-521392403).

glennj:
>... I do agree that we should be doing the `.bash` extension, instead. 

How is that going to affect the CI scripts? I guess we'll have to throw a little extra logic into `bin/validate_exercises`

kotp:
> > ... I do agree that we should be doing the `.bash` extension, instead.
> 
> How is that going to affect the CI scripts? I guess we'll have to throw a little extra logic into `bin/validate_exercises`

No direct answer to this, but no (long term) logic.  All should be `.bash` in my opinion.

MichaelDimmitt:
Really good discussion here 🔥

another thing worth noting, 
diffing is easy to attain in bats

by changing
```bash
@test "a name given" {
  [[ $BATS_RUN_SKIPPED == true  ]] || skip
  run two_fer Alice
  [[ $status -eq 0 ]]
  [[ $output == "One for Alice, one for me." ]]
}
```
```bash
printDiff(){
  echo output;
  echo "$output";
  echo ;
  echo expected;
  echo "$expected;
}

@test "a name given" {
  [[ $BATS_RUN_SKIPPED == true  ]] || skip
  run two_fer Alice
  expected=$(cat <<-END
One for Alice, one for me.
END
  )
  printDiff
  [[ $status -eq 0 ]]
  [[ "$output" ==  "$expected" ]]
}
```

MichaelDimmitt (approved):

ErikSchierboom:
And as a side note: the test runner should also be updated. It might even be necessary to support both formats to not break old exercises. Did we decide how to handle such a situation @iHiD ?

IsaacG:
This PR has seen no activity in 6 months. Is it still being used or should it be closed out?

kotp:
Leave it be.  It is draft and still has a place.

IsaacG:
This PR has seen no activity in 4 years. Closing it out. If we want to revisit this, we can always reopen/recreate it.

PR #376, opened 2019-08-14
list-ops: create exercise

glennj:
---

Reviewer Resources:


exercises/list-ops/list_ops_test.sh
```diff
@@ -0,0 +1,253 @@
+#!/usr/bin/env bash
+
+readonly EXERCISE_VERSION=2.4.0
```
glennj: It might be worthwhile to add the version into the test file itself.
Although the exercism site uses the .meta/version number to indicate when a solution is out of date, 
making the version visible may benefit mentors.

We'd want to add a check into the bin/canonical_data_check to enforce this version equals the .meta/version.

exercises/list-ops/list_ops_test.sh
```diff
@@ -0,0 +1,253 @@
...
+        source list_ops.sh
+        l1=()
+        l2=()
+        list::append l1 "${l2[@]}"
+        (( ${#l1[@]} == 0 )) || exit 1
+        echo "${l1[*]}"
+    '
+    [[ $status -eq 0 ]]
+    [[ $output == "" ]]
+}
```
glennj: The test is a bit of a Frankenmonster: the usual bats tests following the `run`, but also a test inside the inline bash script. I'm not entirely sure I'm happy with it. The actual function call may be too buried.

exercises/list-ops/list_ops_test.sh
guygastineau: I am not opposed to this idea.  I actually tried to make this happen in an earlier version of our tests, but the idea was shot down by the other mentors ("it had no place in the tests as it was of no concern to the students")

I am of the opinion that it is of concern to the students.  Especially now that students can upgrade their exercises to the newest version there is even visibility of the version numbers for the students.

That is to say.. I support this idea.  Please make an issue on the topic, so we can organize integrating this into the track :smile:

guygastineau: I don't think it needs to be a variable though.  A comment would be fine.

exercises/list-ops/list_ops_test.sh
guygastineau: I don't mind it too much, but I see your point.  We really have to go out of our way to get this to work in bash.

glennj: Fixed it with a setup function, and the realization that every line of a test case is an assertion (non-zero exit status)

kotp (approved):

exercises/list-ops/list_ops.sh
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
+
+if [[ ${BASH_VERSINFO[0]} -lt 4 ]]; then
+    echo "This library requires at least bash version 4" >&2
+    return 4
+fi
+
+# Due to inherent bash limitations around word splitting and globbing,
+# functions that are intended to *return a list* are instead required to
+# receive an nameref parameter, the name of an array variable that will be
```
kotp, line 10:
```suggestion
# receive a nameref parameter, the name of an array variable that will be
```

glennj: Before merging, I'll need to rebase and add add the other exercise accoutrements.

glennj:
Before merging, I'll need to rebase and add add the other exercise accoutrements.

exercises/practice/list-ops/README.md
```diff
@@ -0,0 +1,91 @@
...
+* `append` (*given two lists, add all items in the second list to the end of the first list*);
+* `concatenate` (*given a series of lists, combine all items in all lists into one flattened list*);
+* `filter` (*given a predicate and a list, return the list of all items for which `predicate(item)` is True*);
+* `length` (*given a list, return the total number of items within it*);
+* `map` (*given a function and a list, return the list of the results of applying `function(item)` on all items*);
+* `foldl` (*given a function, a list, and initial accumulator, fold (reduce) each item into the accumulator from the left using `function(accumulator, item)`*);
+* `foldr` (*given a function, a list, and an initial accumulator, fold (reduce) each item into the accumulator from the right using `function(item, accumulator)`*);
+* `reverse` (*given a list, return a list with all the original items, but in reversed order*);
+
+
```
glennj, line 22: Is it necessary to talk about this being a library exercise? Will it be obvious enough from the test code?

kotp: If it is the first such type of exercise, it might be a good place to do so.  So "obvious" to some yes, but not necessarily to others.  First time library exercise is a great place to talk about what is being introduced.

exercises/practice/list-ops/.docs/instructions.append.md
```diff
@@ -0,0 +1,28 @@
+# Library of Functions
+
+This is the first exercise we've seen where the solution we're writing
+is not a "main" script. We're writing a library to be "source"d into
+other scripts that will invoke our functions.
```
glennj, line 5: Is this sufficient to talk about "libraries"?

kotp: I think it is, and if not, we will find out about it from students.

PR #379, opened 2019-08-19
Resistor-color-duo: add test with 3 colors, and third color is ignored.

glennj:
Based on problem-specifications pull request 1569
https://github.com/exercism/problem-specifications/pull/1569

Recerting the PR after pooching up the git history.

---

Reviewer Resources:


guygastineau:
Cool.  We need to wait until it is merged there.

Thanks.

Issue #382, opened 2019-08-26
Issue with test on Sieve exercise for Bash track.


DERPSQUID9:
My script for Sieve on the Bash track appears to output the same values as expected by the test script when I run it on its own, but it only ever passes the first two tests no matter what I do.

iHiD:
FYI, I've moved this to exercism/bash

IsaacG:
So ... bats `run` changes the IFS.

```
@test 'basic' {
  n=$IFS
  [[ $n = "$IFS" ]]
  run bash -c 'echo "${a[@]}"'
  [[ $n = "$IFS" ]]
}
```

^^ fails on the second test. But passes the first.

```
@test 'basic' {
  echo Before
  echo "$IFS" | od -c

  run bash -c 'echo "${a[@]}"'
  echo After
  echo "$IFS" | od -c

  false
}
```

prints

```
   Before
   0000000      \t  \n  \n
   0000004
   After
   0000000  \n  \n
   0000002
```

So the IFS is being set to `\n` where it generally is `space-tab-newline`. Which means the output gets newline-separated vs space.

IsaacG:
There's a bug in the test tool [here on L61](https://github.com/sstephenson/bats/blob/03608115df2071fff4eaaff1605768c275e5f81f/libexec/bats-exec-test#L61). Rather than assigning IFS for the one line where they want to split the output on newline, they are actually permanently assigning a new value to IFS in the format of `a=1 b=2` which assigns two vars.

I'll submit an upstream patch.

IsaacG:
Fun fact, that bats repo has: Latest commit 0360811 on Feb 19, 2016.

It might not get fixed. Work around might require us setting IFS=$' \t\n' in the test code after doing a `run`

glennj:
PR https://github.com/exercism/bash/pull/383 takes another approach: don't use arrays for expected data in the tests.

glennj:
That PR has been merged. You can navigate to your sieve solution and update to the latest test version. Re-download and verify your working code.

guygastineau:
@DERPSQUID9 can you confirm your solution is working now?

DERPSQUID9:
@guygastineau I sure can!

PR #383, opened 2019-09-02
Bats fix arrays in tests

glennj:
The `bats` that gets installed as a package on ubuntu alters IFS while the tests are running. This breaks the tests for `series` and `sieve`.

It seems the simplest solution is to stop using arrays in the tests.

---

Reviewer Resources:


glennj:
I'm surprised this didn't break on CI. I'm using bats-core on my mac: I ended up installing a ubuntu VM to reproduce.

kotp:
Thanks for the work on this.

guygastineau:
Sweet.  Did this fix #382 ?

glennj:
Yes it did.

PR #387, opened 2019-09-09
Add WSL to installation guide for Windows

CalebRollins:


CalebRollins:
I'm not an expert in this area, but in my experience WSL is much nicer and I think it's the way most people are going. I'll change it.

glennj:
I know this is off-topic for this PR, but it's tangentially related: for Mac users, do we want to recommend using Homebrew to install an up-to-date version of bash?

guygastineau:
@CalebRollins qell, it seems us non-WSL-experts are in agreement that it is better than cygwin 🤣

I suggest you go ahead and put it first then 😁

@glennj I was thinking about that earlier.
Specifically regarding the library testing I wanted us to put a disclaimer that this track is designed specifically for bash 3.2 and above.  Then, there could be a comment as well suggesting that folks should consider getting the newest bash available.

I guess I mean to say:
- we agreed that version 3.2+ should be assumed, but not to force newer versions.
- I think it is appropriate for us to encourage people to use the most up to date bash they can get on their OS.

guygastineau:
Maybe I misread the diff and you always had it first 🤣

PR #390, opened 2019-09-13
Hamming: streamline error messages

glennj:
This change consolidates the different error messages into 1 while still keeping all the test scenarios.

Looking at the history of the canonical-data, there was an edge case for recursive solutions in Haskell where left/right empty input was not picked up. They chose to use specific error messages for these situations.

They don't really add any value for learning bash.

---

Reviewer Resources:


guygastineau:
Thanks

PR #392, opened 2019-09-30
Expanding the list of alternatives for installing bash on Windows.

glennj:
Plus, tweak the Linux wording a bit.

---

Reviewer Resources:

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,9 +1,10 @@
 ## Linux
 
-Bash, as the GNU shell, will be found on any GNU/Linux system. It is usually found on any BSD or Unix based system.
+Bash, as the GNU shell, will be found on any GNU/Linux system. 
+BSD and other UNIX derivatives will likely require installation by the user.
+If you are on BSD, AIX, or a Linux that doesn't come with bash please consult your operating system's resources regarding installation of non-core packages.
 
-In the graphical interface of your choice, look for an application named "Terminal".
-Alternatively, Ctrl+Alt+F2 should take you to a system console (Ctrl+Alt+F1 or F7 to return to your Graphical desktop).
+In the graphical interface of your choice, look for an application named "Terminal": it might be found under an "Administrative/System Tools" section.
 
 To verify the version of Bash available, from the command line or terminal, run:
 
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,6 +1,9 @@
-Bash is usually found on any Linux, BSD, or Unix based systems.
+## Linux
+
+Bash, as the GNU shell, will be found on any GNU/Linux system. It is usually found on any BSD or Unix based system.
+
 In the graphical interface of your choice, look for an application named "Terminal".
-Alternatively, Ctrl+Alt+F2 should take you to a shell login (Ctrl+Alt+F1 or F7 to return to Graphical mode).
+Alternatively, Ctrl+Alt+F2 should take you to a system console (Ctrl+Alt+F1 or F7 to return to your Graphical desktop).
 
 To verify the version of Bash available, from the command line or terminal, run:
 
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -32,13 +35,25 @@ This is free software; you are free to change and redistribute it.
...
 
-This should output some terse information about the version of bash you are running.
 If it does not, you may not have bash installed.
 You should install it using your distributions package manager.
 
-For Windows, you may need to install [WSL](https://docs.microsoft.com/en-us/windows/wsl/install-win10). Alternatively, you can install [Cygwin](https://cygwin.com/).
+## Windows
+
+For Windows 10, you can install [WSL](https://docs.microsoft.com/en-us/windows/wsl/about). 
+This installs a full (well, the non-GUI components) Linux distribution that runs natively in Windows.
```
guygastineau: :rofl: yeah it is probably good to tell windows users not to expect a graphical environment.

docs/INSTALLATION.md
```diff
@@ -32,13 +35,25 @@ This is free software; you are free to change and redistribute it.
...
+
+Alternatively, you can install:
+
+* [Cygwin](https://cygwin.com/)
+* [git for windows](https://gitforwindows.org/)
+* [MinGW](http://mingw.org/)
+* a full Linux distribution in [a virtual machine](https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/about/)
```
guygastineau: Very nice comprehensive list thank you :heart_eyes:

docs/INSTALLATION.md
```diff
@@ -1,6 +1,9 @@
-Bash is usually found on any Linux, BSD, or Unix based systems.
+## Linux
+
+Bash, as the GNU shell, will be found on any GNU/Linux system. It is usually found on any BSD or Unix based system.
```
guygastineau: I never noticed this assumption about `BSD` before. I am glad that this PR made me see it.

For the record I have never gotten `bash` out of the box on a `BSD` system.  I have always either installed the pkg or used ports to install it from scratch.

I guess we might need to update how we talk about `BSD` here too...

glennj: Aside from MacOS, I have no BSD experience.

guygastineau: I also think thhat `AIX` requires installing bash after the fact too, but I don't remember how (I have only consulted for projects using `AIX`).

@kotp might know more about the `AIX` stuff.

Concerning `BSD` I don't know how many `BSD` variants we want to mention explicitly...
Maybe we could just note that `bash` probably needs to be installed by the user, and then link to the install instruction pages for `FreeBSD`, `NetBSD`, `OpenBSD`, and `Dragonfly BSD`?

guygastineau: Just to get this PR through (and without expanding it's scope) I would be fine with a short section for BSD that says:
```
On BSD systems bash will likely need to be installed by the user.
Please refer to your operating system's resources concerning installation
```
Then I can make a PR that gives better details tonight or tomorrow?

In the meantime I doubt we are going to lose any new students using BSD.  In my experience BSD users don't need much hand-holding plus some of them might have a misguided sense that `csh` is better :rofl:

Kidding aside, there should be good instructions for BSD, but I would prefer if it were at least accurate in the mean time.

glennj: Hmm, rather than list *any* BSD or *nix variants, maybe we just have a blanket statement to consult the community for your niche OS (although without the condescension).

guygastineau: Cool.
So, for the time being, lets try:
```
Bash is usually found on any linux.

BSD and other UNIX derivatives will likely require installation by the user.
If you are on BSD, AIX, or a linux that doesn't come with Bash please consult your operating system's resources regarding installation of non-core packages
```

Or something to that effect.  I am flexible regarding the wording.

guygastineau: Then, like I said above, I might do some more research, and then provide clearer instructions or links for the most popular OS in that group.

kotp: > I also think that `AIX` requires installing bash after the fact too, but I don't remember how (I have only consulted for projects using `AIX`).
> 
> @kotp might know more about the `AIX` stuff.

I believe it is `bsh` Bourne Shell on AIX recently.  Bash isn't (as far as I remember) generally installed by default.  They [have these bash rpm packages](http://ftp.software.ibm.com/aix/freeSoftware/aixtoolbox/RPMS/ppc/bash/) available.

docs/INSTALLATION.md
```diff
@@ -1,6 +1,9 @@
...
+
 In the graphical interface of your choice, look for an application named "Terminal".
-Alternatively, Ctrl+Alt+F2 should take you to a shell login (Ctrl+Alt+F1 or F7 to return to Graphical mode).
+Alternatively, Ctrl+Alt+F2 should take you to a system console (Ctrl+Alt+F1 or F7 to return to your Graphical desktop).
```
kotp: I am not sure that this is good advice.  For those that are familiar with their terminals, they will already know this, and the issues that can happen.  Since virtual terminal 7 is not always the thing, anyway, we may end up with folks needing to reboot since they are now in an unfamiliar state.  If we were there synchronously with the people that might do this, we would be able to talk them back out of this, if their virtual terminals were different.

There is just too many ways that this could be configured, and those that know won't have any problems with this, while the people that we are trying to instruct (the one that aren't familiar) are very likely to be seeing this for the first time, and they aren't the ones we really want to be doing this.

guygastineau:
Thank you.

PR #393, opened 2019-09-30
expand Mac installation section

glennj:
Homebrew instructions for Mac.

This should not conflict with my other PR expanding Windows instructions.

---

Reviewer Resources:

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -43,3 +43,28 @@ For Apple OSX:
 * open your Applications folder,
 * open the Utilities folder,
 * finally open the `Terminal` application.
+
+Note that the bash version installed on MacOS is very old, version 3.2. 
+This is due to (i) newer bash versions being licensed under the GPLv3 and (ii) Apple not wanting to be forced to comply with the terms of that license.
+Additionally, Apple has signaled that bash will be going away from MacOS: the default login shell was recently [changed from bash to zsh](https://support.apple.com/en-us/HT208050).
+
+To obtain an up-to-date bash on your Mac:
+
+1. install [Homebrew](https://brew.sh)
+1. open a Terminal and enter: `brew install bash`
+
+This installs an up-to-date bash as (I believe) `/usr/local/bin/bash` (so that the system's default bash is not overwritten -- other tools may rely on /bin/bash). 
+
+If you want homebrew bash to be your default bash:
+
+1. add to your ~/.bash_profile (or ~/.profile) file:
+
+        [[ ":$PATH:" == *":/usr/local/bin:"* ]] || PATH=/usr/local/bin:$PATH
+
+1. add it to the list of allowed login shells:
+
+        echo /usr/local/bin/bash | sudo tee -a /etc/shells
+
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -43,3 +43,28 @@ For Apple OSX:
...
+
+1. add to your ~/.bash_profile (or ~/.profile) file:
+
+        [[ ":$PATH:" == *":/usr/local/bin:"* ]] || PATH=/usr/local/bin:$PATH
+
+1. add it to the list of allowed login shells:
+
+        echo /usr/local/bin/bash | sudo tee -a /etc/shells
+
+1. change your login shell:
```
guygastineau: The only confusion here would be that on up to date Macs this would change their default shell from `zsh` to `bash`, but I don't think we should have a bunch of clutter to describe the differences.

Maybe it could say:
```
If you want bash as your default login shell then:
  chsh -s /usr/local/bin/bash
```

Or something along those lines.  I am interested in your ideas about this too.  I don't want to mislead students, but we also aren't teaching an class on how to configure `*nix` operating systems (well sort of)

guygastineau: IMO, this runs the risk of users with older macs opening a terminal without doing this and thinking it was already set up.

docs/INSTALLATION.md
```diff
@@ -43,3 +43,28 @@ For Apple OSX:
...
+Additionally, Apple has signaled that bash will be going away from MacOS: the default login shell was recently [changed from bash to zsh](https://support.apple.com/en-us/HT208050).
+
+To obtain an up-to-date bash on your Mac:
+
+1. install [Homebrew](https://brew.sh)
+1. open a Terminal and enter: `brew install bash`
+
+This installs an up-to-date bash as (I believe) `/usr/local/bin/bash` (so that the system's default bash is not overwritten -- other tools may rely on /bin/bash). 
+
+If you want homebrew bash to be your default bash:
```
guygastineau: Maybe this is why I have a problem with the earlier comment.

Line `68` tells how to set the default shell (which might not be bash).

I think I have seen before how to make the system prefer the brew installed bash over the default without changing the login shell.  Does anybody know where to find that documentation?

guygastineau:
This [post](https://scriptingosx.com/2019/02/install-bash-5-on-macos/) suggests that Mac puts `/usr/local/bin` before `/bin` in `PATH`, so users who are using zsh shouldn't have to do anything for bash to work (and to override the old bash if it is present at all).

So the line before the last section should probably be:
```
If you want bash version 5.x as your default login shell then please follow these instructions.

guygastineau:
Let me know if that last suggestion makes sense @glennj 
Or if it is wrong entirely...

glennj:
it does make sense. There are 2 things that I'm *trying* to express: 1) which bash binary you use when you type `bash` (determined by the PATH); and 2) user's login shell.

I think now that it's not appropriate to talk about the login shell in this installation document. But I can be more clear about what I mean by "default bash". I'll submit a new edit.

PR #397, opened 2019-10-03
reverse-string: add test that exposes unquoted variable problems

glennj:
Addressing same problem as PR #377 

---

Reviewer Resources:


guygastineau:
Is this our own test? (ie. not from `problem-specifications`)

glennj:
> Is this our own test? (ie. not from `problem-specifications`)

Yes. It comes from seeing the same mistake over and over while mentoring. 

I would not think this kind of problem exists in other languages, so I don't think it's worth submitting to the canonical data.

guygastineau:
Right, but we don't get to increment the version past the `canonical-data` version I don't believe

guygastineau:
The Dashboard doesn't seem to care, but what happens if the `canonical-data.json` is upgraded? We wont be able to tell that our exercise is behind the new information in the `problem-specifications` repository.

guygastineau:
BTW, sorry I didn't see your response earlier.  For the record I think this is a great test to add to this exercise for our track! :nerd_face:

glennj:
I've done it before, see https://tracks.exercism.io/bash/master (word-count, for example) -- the major and minor versions are the same, just the patch number is incremented.

guygastineau:
Right, sometimes test semantics are changed and they just increment the patch number, so we will still be getting out of sync.

Unless I am missing something here.

PR #398, opened 2019-10-03
Reverse String: Revert Version

guygastineau:
If the version goes past the canonical-data version then we can't tell when it gets behind after the next upstream change.

[canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/reverse-string/canonical-data.json)

---

Reviewer Resources:


glennj:
From what I've seen so far,  changes to canonical data increment the minor version, not the patch version.

guygastineau:
From the `problem-specifications` `README.md`
Test Data Versioning

Test data should be versioned according to Semantic Version 2.0, which is defined as follows:

    Given a version number MAJOR.MINOR.PATCH, increment the:

    MAJOR version when you make incompatible API changes, MINOR version when you add functionality in a backwards-compatible manner, and PATCH version when you make backwards-compatible bug fixes. Additional labels for pre-release and build metadata are available as extensions to the MAJOR.MINOR.PATCH format.

MAJOR version changes

The MAJOR version should be changed when the test suite is modified in a fundamentally incompatible way.

There are examples of changes requiring a MAJOR version change:

    A new "property" (test type).
    Renaming a "property".
    Insertion, deletion or renaming of keys in the test data object.
    Changing the type of one of the test data keys.

MAJOR changes should be expected to break even well-behaved test generators.
MINOR version changes

The MINOR version should change when you add functionality in a backwards-compatible manner, make non-breaking changes that alter the meaning of the test suite, make previously passing solutions possibly fail, or failing solutions fail at a different spot.

There are examples of changes requiring a MINOR version change:

    Adding or deleting test cases.
    Changing the test cases inputs and/or outputs.
    Changing the test cases ordering.

MINOR changes would never break well-designed test generators, because the test-generation logic remains exactly the same.
PATCH version changes

The PATCH version should change when you make backwards-compatible bug fixes or whenever the meaning of the tests does not change.

There are examples of changes requiring a PATCH version change:

    Regrouping/"Renesting" test cases without changing test case ordering.
    Changing descriptions or comments.
    Changing keys' ordering or formatting (would result in an equivalent JSON file).

PATCH changes would never break well-designed test generators, because the test data remains exactly the same.

guygastineau:
I thought that exercism detected that there was a new commit, and provided the ability to update based on that.  Maybe I'm wrong, but I know that (at least in the past) the student exercise is pinned to a commit hash not to a version number.

guygastineau:
It might be true that the patch version changes wont mean any semantic changes for our tests.  In that case I guess I am fine with us changing it to reflect our local changes, but it makes me uneasy to deviate from the canonical versions

guygastineau:
I suppose I am just a much more comfortable if we make our own versions for augmentations we make to the tests.  It would be another thing to manage, but it would make what is going on a lot clearer.

glennj:
Well, would the version number `1.2.0.1` be OK?

guygastineau:
I think you are on to something there.

We need to make sure that (https://github.com/exercism/bash/blob/master/bin/canonical_data_check) will still work. We should also make sure that it won't mess with the dashboard's ability to recognize the first 3 numbers.

I think a good compromise could be:
- put version in front of the canonical version with a space. In the above script we could grep for version and print the second field of that line.
- then the second line could be prefaced with local_patch followed by a patch number for book keeping on our custom changes.

What do you think? If your suggestion can work with the script and the dashboard, then I think it is fine, but I still like the clean separation of the compromise a little more.

glennj:
> Well, would the version number `1.2.0.1` be OK?

Had a look at the track maint. dashboard source, and a 4-component version will *not* work. Ref  https://github.com/exercism/tracks-maintenance-dashboard/blob/master/src/hooks/useRemoteVersion.ts#L144

It appears that for tracks that use .meta/version as the versioning scheme, the dashboard just reads the contents of that file.

We are probably in territory where we want to put the version into the _test file:

* keep the contents of .meta/version identical to the canonical version, 
* add some commentary into the _test file with 
    1. the canonical version number and
    1.  any comments about bash-only tests.
* yes this puts the version number in 2 places, but updating the bin/canonical_data_check can alert when they're not updated together.

I'd be willing to submit a large PR to do this.

glennj:
> > Well, would the version number `1.2.0.1` be OK?
> 
> Had a look at the track maint. dashboard source, and a 4-component version will _not_ work. Ref https://github.com/exercism/tracks-maintenance-dashboard/blob/master/src/hooks/useRemoteVersion.ts#L144

Looking closer, it should be OK for the dashboard: that regex is not anchored.

guygastineau:
> Although this still does leave us with the problem where we add bash-only tests to an existing _test file: exercise authors won't be notified that they need to update their solutions.

I don't think that is true.

> Looking closer, it should be OK for the dashboard: that regex is not anchored.

Which solution are you suggesting here?

glennj:
> > Although this still does leave us with the problem where we add bash-only tests to an existing _test file: exercise authors won't be notified that they need to update their solutions.
> 
> I don't think that is true.

I see you're right. It appears to be keyed on git commit ids.

> > Looking closer, it should be OK for the dashboard: that regex is not anchored.
> 
> Which solution are you suggesting here?

After reading a bit about semantic versioning, we need to keep the version as MAJOR.MINOR.PATCH -- so I'm suggesting my solution: https://github.com/exercism/bash/pull/398#issuecomment-538466021

guygastineau:
Yeah, I found out from iHiD that updates for exercism are only triggered when the test file itself is updated, so if we put our local version info in the test file, then we could increment it for things like subtle differences in a readme or something, and that would trigger an update for the students.  It sounds like a wini/win 😁

guygastineau:
Also, a large PR for this would be great 😀

glennj:
This PR can be expanded to fix these other exercises which are a patch increment above canonical: acronym, word-count, proverb, pig-latin, series, sieve

glennj:
Or, how about we close this PR and I'll take care of them

guygastineau:
Sure, you can handle the rest if you want.

How are you wanting to put the local patch info in the tests files? I can just add the new convention here for `reverse-string`

glennj:
Ref PR #399

guygastineau:
Thanks

PR #400, opened 2019-10-09
hamming: add test to expose == as a pattern matching operator

glennj:
Realized need for this test in a mentoring session.

---

Reviewer Resources:


kotp:
Good find.

PR #402, opened 2019-10-21
simple-cipher: issue with extglob in test file

glennj:
Running with bats-core and bash 3.2, the extended pattern in the first test errored.

---

Reviewer Resources:


guygastineau:
Thank you.

PR #404, opened 2019-11-09
Added bats install instructions for windows / cygwin

siad007:
Added `bats` install instructions for windows under Cygwin / MINGW64

---

Reviewer Resources:

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -28,8 +28,7 @@ $ git clone https://github.com/sstephenson/bats.git
 $ cd bats
 $ ./install.sh $HOME
 ```
-If you discover an error like `cp: cannot create symbolic link '/c/Users/XXX/bin/bats': No such file or directory`,
-you have to copy the `bin/bats/libexec/` folder content to `/c/Users/XXX/bin/` manually.
+If you discover an error like `cp: cannot create symbolic link '${HOME}/bin/bats': No such file or directory`, you have to copy the `bin/bats/libexec/` folder content to `${HOME}/bin/` manually.
 
 ---
 
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -24,11 +24,11 @@ sudo yum install bats
 
 ### For Windows (MINGW64/Cygwin)
 ```
-$ git clone https://github.com/sstephenson/bats.git
+$ git clone https://github.com/bats-core/bats-core.git
 $ cd bats
 $ ./install.sh $HOME
 ```
-If you discover an error like `cp: cannot create symbolic link '${HOME}/bin/bats': No such file or directory`, you have to copy the `bin/bats/libexec/` folder content to `${HOME}/bin/` manually.
+Note: When you are using the outdated `https://github.com/sstephenson/bats.git` and you discover an error like `cp: cannot create symbolic link '${HOME}/bin/bats': No such file or directory`, you have to copy the `bin/bats/libexec/` folder content to `${HOME}/bin/` manually.
 
 ---
 
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,5 +1,6 @@
 As there isn't much of a bash ecosystem, there also isn't really a defacto leader in the bash testing area. For these examples we are using [bats](https://github.com/sstephenson/bats). You should be able to install it from your favorite package manager, on OS X with homebrew this would look something like this:
 
+### For Mac (brew)
 ```
 $ brew install bats
 ==> Downloading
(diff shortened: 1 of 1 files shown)


guygastineau:
This looks reasonable, but I have never used cygwin.  It would be great if someone else with experience in this could review this PR.

@siad007 Thank you for your work.  Sorry you have had to wait so long.

docs/TESTS.md
```diff
@@ -11,14 +12,25 @@ https://codeload.github.com/sstephenson/bats/tar.gz/v0.4.0
...
 ```
 
+### For Windows (MINGW64/Cygwin)
+```
+$ git clone https://github.com/sstephenson/bats.git
+$ cd bats
+$ ./install.sh $HOME
+```
+If you discover an error like `cp: cannot create symbolic link '/c/Users/XXX/bin/bats': No such file or directory`,
+you have to copy the `bin/bats/libexec/` folder content to `/c/Users/XXX/bin/` manually.
```
kotp: To keep this file consistent, this should be on the same line, similar to line 1.

docs/TESTS.md
```diff
@@ -11,14 +12,25 @@ https://codeload.github.com/sstephenson/bats/tar.gz/v0.4.0
...
 ```
 sudo yum install bats  
 ```
 
+### For Windows (MINGW64/Cygwin)
+```
+$ git clone https://github.com/sstephenson/bats.git
+$ cd bats
+$ ./install.sh $HOME
+```
```
kotp: If `$HOME` is available, line 32 could reference `$HOME` as well, in order to remove any ambiguity, given the `XXX` that is shown there in place of, I assume, the login name.

siad007:
> 
> 
> Style things to change.

Done.

docs/TESTS.md
```diff
@@ -11,14 +12,24 @@ https://codeload.github.com/sstephenson/bats/tar.gz/v0.4.0
...
 
-For Red Hat, Scientific Linux, and CentOS 6 or later bats is found in the EPEL repository.  
+### For Red Hat, Scientific Linux, and CentOS 6 or later bats is found in the EPEL repository.  
 ```
 sudo yum install bats  
 ```
 
+### For Windows (MINGW64/Cygwin)
+```
+$ git clone https://github.com/sstephenson/bats.git
```
guygastineau: That repository is deprecated. No commits since 2016.

Everyone should be using: `https://github.com/bats-core/bats-core.git`

guygastineau:
@kotp thank you for the astute observations.  I believe your requests have been satisfied now.

@siad007 Excellent work satisfying @kotp's requests.  While reviewing that work I realized that the instructions were pointed at a deprecated repository.  After that is fixed I believe this PR will be ready to go ;)

siad007:
> 
> 
> @kotp thank you for the astute observations. I believe your requests have been satisfied now.
> 
> @siad007 Excellent work satisfying @kotp's requests. While reviewing that work I realized that the instructions were pointed at a deprecated repository. After that is fixed I believe this PR will be ready to go ;)

So should I change all occurrences with the new repo? I mean all other os installs are using the same "old" https://github.com/sstephenson/bats.git in the documentation.

kotp:
> @guygastineau So should I change all occurrences with the new repo? I mean all other os installs are using the same "old" https://github.com/sstephenson/bats.git in the documentation.

Since this PR is titled and concerning Windows installation, perhaps not.  You could create a new pull request that address the correction over all, and leave this one to address the instructions for Windows and Cygwin.

guygastineau:
Well, the only other place that I see reference to he `stephenson/bats` repo is in the expected output from `brew` on `OS X`.  I don't currently have a mac, so I can't test that, but if that is what `brew` uses, then there is not much we can do.  Maybe there is a way to make a PR to `brew` to update where it gets that package.  That should happen, but we can't really change what that output is until it is reality.  Does that make sense?

TL;DR They aren't manual instructions for `OS X`. Rather it appears in expected output from `brew`.  Whenever `brew` starts using the maintained repo we can change it here.

PS. Perhaps we should suggest that mac users install from the git source manually if they want the maintained version of `BATS`?  Only as an alternative?  It still wouldn't belong in this PR, but we should consider this.

docs/TESTS.md
```diff
@@ -11,14 +12,24 @@ https://codeload.github.com/sstephenson/bats/tar.gz/v0.4.0
...
 sudo yum install bats  
 ```
 
+### For Windows (MINGW64/Cygwin)
+```
+$ git clone https://github.com/bats-core/bats-core.git
+$ cd bats
+$ ./install.sh $HOME
+```
+Note: When you are using the outdated `https://github.com/sstephenson/bats.git` and you discover an error like `cp: cannot create symbolic link '${HOME}/bin/bats': No such file or directory`, you have to copy the `bin/bats/libexec/` folder content to `${HOME}/bin/` manually.
```
guygastineau: Cool, so this problem doesn't exist with `bats-core`?

guygastineau:
Awesome work, @siad007 .

 @kotp I think your change request has been satisfied too?

glennj:
> Well, the only other place that I see reference to he `stephenson/bats` repo is in the expected output from `brew` on `OS X`. I don't currently have a mac, so I can't test that, but if that is what `brew` uses, then there is not much we can do. 

brew has both the stepthenson/bats (`brew install bats`) and the bats-core/bats (`brew install bats-core`). I'll create a PR for the change to the test documentation.

guygastineau:
Thank you @siad007  :confetti_ball:

PR #406, opened 2019-11-25
Fix topic names

bsoyka:
After viewing an [analysis of the topics on the Bash track](https://tracks.exercism.io/bash/master/topics), I noticed a lot of topics that weren't listed in [`exercism/problem-specifications/TOPICS.txt`](https://github.com/exercism/problem-specifications/blob/master/TOPICS.txt). This PR replaces most of those with valid topic names.

I'm always completely open to any changes.

---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -21,7 +21,7 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-        "control_flow_conditionals",
+        "conditionals",
         "strings",
         "text_formatting"
       ]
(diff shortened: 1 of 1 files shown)


glennj:
I have a suggestion: https://docs.google.com/spreadsheets/d/14c1oovaMJeHvxBWpCk6BqX_XYqNBgLHfQ1WKsM-zye8/edit?usp=sharing

config.json
```diff
@@ -171,9 +172,9 @@
       "unlocked_by": "pangram",
       "difficulty": 2,
       "topics": [
-        "associative_arrays",
```
guygastineau: pangram doesn't even need arrays really.  I am interested in opinions about dropping `arrays` for `pangram`...

guygastineau:
@glennj are we waiting on an implementation of your suggestions?

Issue #409, opened 2019-11-30
Markdown exercise: tests on mac are fail but on ubuntu ok


b10s:
how I can see it on my mac:
![photo_2019-11-30_20-18-17](https://user-images.githubusercontent.com/4618288/69899782-95ff5f80-13ae-11ea-9c86-6970dcc8c6d9.jpg)

b10s:
Just have identified the part which works incorrect in my OSX bash:
`str='## abcd'; expr "$str" : "#\+"`

<img width="461" alt="Screen Shot 2019-12-02 at 17 48 56" src="https://user-images.githubusercontent.com/4618288/69945026-10092300-152c-11ea-90a7-7b4483908c8a.png">

glennj:
It seems the GNU expr considers `#\+` a valid "basic regular expression", but the BSD expr does not:
```bash
$ str='## abcd'; expr "$str" : "#\+"
2
$ /bin/expr "$str" : "#\+"
0
$ /bin/expr "$str" : '#\+'
0
$ /bin/expr "$str" : '#\\+'
0
$ /bin/expr "$str" : '#*'
2
$ /bin/expr "$str" : '#{1,}'
0
$ /bin/expr "$str" : '#\{1,\}'
2
```
This *appears* to be buggy behaviour, but I can't find a reference in the BSD expr man page describing the basic regex syntax being used.

guygastineau:
This issue is now resolved?

glennj:
Right, fix has been merged.

PR #410, opened 2019-12-02
markdown exercise: pre-refactored code does not pass tests on OSX

glennj:
ref issue https://github.com/exercism/bash/issues/409

---

Reviewer Resources:

### DIFF
--- a/exercises/markdown/markdown.sh
+++ b/exercises/markdown/markdown.sh
@@ -36,7 +36,7 @@ done
         inside_a_list=no
     fi
 
-            n=`expr "$line" : "#\+"`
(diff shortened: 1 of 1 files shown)


exercises/markdown/markdown.sh
```diff
@@ -36,7 +36,7 @@ done
         inside_a_list=no
     fi
 
-            n=`expr "$line" : "#\+"`
+            n=`expr "$line" : "#\{1,\}"`
```
guygastineau: So this fixes it?

glennj: Yessir.

guygastineau: Thank you

PR #411, opened 2019-12-03
hamming test suite: test name does not display correctly

glennj:
Slight tweak to make the test name appear correctly during a bats run.

---

Reviewer Resources:


guygastineau:
This looks good.  Also doesn't look like it is going to cause any of the older PR's to need rebasing, so I will go ahead and merge this.

As always thank you for the good work @glennj

PR #414, opened 2019-12-23
add .meta/config.json files to each exercise for the test runner

guygastineau:
This almost worked for all of them, but alas there are some edge cases that make it difficult to automate grabbing the run lines when there are multiple of them per test case. In such case my automated solution favored failing.

I could switch to an algorithm that builds along the way, but I still don't know exactly what we should do for the test cmd in such cases (since it is supposed to be a string not a list).

I will probably just have it use the first run string in cases with multiple per case.
Anyway, I figured I would let this draft go up here, but I should have all of them done tomorrow.

Closes #414 

---

Reviewer Resources:

### DIFF
--- a/exercises/dnd-character/.meta/config.json
+++ b/exercises/dnd-character/.meta/config.json
@@ -73,7 +73,7 @@
     },
     {
       "name": "validate ability and hitpoint range",
-      "cmd": "< <( run bash dnd_character.sh generate )"
+      "cmd": "bash dnd_character.sh generate"
     }
   ]
 }
--- a/exercises/dnd-character/.meta/config.json
+++ b/exercises/dnd-character/.meta/config.json
@@ -1,0 +1,79 @@
+{
+  "solution_files": [
+    "dnd_character.sh"
+  ],
+  "tests": [
+    {
+      "name": "ability modifier for score 3 is -4",
+      "cmd": "bash dnd_character.sh modifier 3"
+    },
+    {
+      "name": "ability modifier for score 4 is -3",
+      "cmd": "bash dnd_character.sh modifier 4"
+    },
+    {
+      "name": "ability modifier for score 5 is -3",
+      "cmd": "bash dnd_character.sh modifier 5"
+    },
+    {
+      "name": "ability modifier for score 6 is -2",
+      "cmd": "bash dnd_character.sh modifier 6"
+    },
+    {
+      "name": "ability modifier for score 7 is -2",
+      "cmd": "bash dnd_character.sh modifier 7"
+    },
+    {
+      "name": "ability modifier for score 8 is -1",
+      "cmd": "bash dnd_character.sh modifier 8"
+    },
+    {
+      "name": "ability modifier for score 9 is -1",
+      "cmd": "bash dnd_character.sh modifier 9"
+    },
+    {
+      "name": "ability modifier for score 10 is 0",
+      "cmd": "bash dnd_character.sh modifier 10"
+    },
+    {
+      "name": "ability modifier for score 11 is 0",
+      "cmd": "bash dnd_character.sh modifier 11"
+    },
+    {
+      "name": "ability modifier for score 12 is +1",
+      "cmd": "bash dnd_character.sh modifier 12"
+    },
+    {
+      "name": "ability modifier for score 13 is +1",
+      "cmd": "bash dnd_character.sh modifier 13"
+    },
+    {
+      "name": "ability modifier for score 14 is +2",
+      "cmd": "bash dnd_character.sh modifier 14"
+    },
+    {
+      "name": "ability modifier for score 15 is +2",
+      "cmd": "bash dnd_character.sh modifier 15"
+    },
+    {
+      "name": "ability modifier for score 16 is +3",
+      "cmd": "bash dnd_character.sh modifier 16"
+    },
+    {
+      "name": "ability modifier for score 17 is +3",
+      "cmd": "bash dnd_character.sh modifier 17"
+    },
+    {
+      "name": "ability modifier for score 18 is +4",
+      "cmd": "bash dnd_character.sh modifier 18"
+    },
+    {
+      "name": "generate a character",
+      "cmd": "bash dnd_character.sh generate"
+    },
+    {
+      "name": "validate ability and hitpoint range",
+      "cmd": "< <( run bash dnd_character.sh generate )"
+    }
+  ]
+}
--- a/exercises/ocr-numbers/.meta/config.json
+++ b/exercises/ocr-numbers/.meta/config.json
@@ -1,0 +1,79 @@
+{
+  "solution_files": [
+    "ocr_numbers.sh"
+  ],
+  "tests": [
+    {
+      "name": "No input",
+      "cmd": "bash ocr_numbers.sh"
+    },
+    {
+      "name": "Recognizes 0",
+      "cmd": "bash ocr_numbers.sh << INPUT"
+    },
+    {
+      "name": "Recognizes 1",
+      "cmd": "bash ocr_numbers.sh << INPUT"
+    },
+    {
+      "name": "Unreadable but correctly sized inputs return ?",
+      "cmd": "bash ocr_numbers.sh << INPUT"
+    },
+    {
+      "name": "Input with a number of lines that is not a multiple of four raises an error",
+      "cmd": "bash ocr_numbers.sh << INPUT"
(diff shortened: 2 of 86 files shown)


exercises/dnd-character/.meta/config.json
```diff
@@ -0,0 +1,79 @@
...
+      "name": "ability modifier for score 18 is +4",
+      "cmd": "bash dnd_character.sh modifier 18"
+    },
+    {
+      "name": "generate a character",
+      "cmd": "bash dnd_character.sh generate"
+    },
+    {
+      "name": "validate ability and hitpoint range",
+      "cmd": "< <( run bash dnd_character.sh generate )"
```
guygastineau: I remembered seeing this when things failed for this the first time.

I suppose it is a spot where manual intervention could be a good idea.

exercises/ocr-numbers/.meta/config.json
```diff
@@ -0,0 +1,79 @@
...
+    "ocr_numbers.sh"
+  ],
+  "tests": [
+    {
+      "name": "No input",
+      "cmd": "bash ocr_numbers.sh"
+    },
+    {
+      "name": "Recognizes 0",
+      "cmd": "bash ocr_numbers.sh << INPUT"
```
guygastineau: Hahahaha. This one came out pretty funny!

I wonder what the best thing to do here is.  Basically, I want these changes to go up before I have time to write a sufficiently smart parser for handling multiline values here.  Furthermore, do we actually want multiline values in these files?  We probably need to figure out our position on these things.  Some of the test have some huge inputs.  That is a little worrysome :/

guygastineau:
Hmm, so the couple things I brought up above should be considered before this is merged, but I think this is ready for some conversation.  Mostly about how to deal with 2 inconvenient edge cases here.

I will put the code I used to automate this up on GitHub, and link it here.  I wrote in Haskell, so there is no way that it's source should be held within this repository.  If we think it is a good way to go forward, then we could pull it in like how `configlet` is pulled into the repo.

After seeing what I have put together for it, if people are interested in integrating it somewhat I am interested in getting peoples opinions about names.  Right now it is called `configurator` on my local machine.  That is funny, but it is neither super accurate nor descriptive.  In reality I made a little script to control, because I wanted more `bash` involved, and I was excited to use the nifty `interact` function from Haskell to encapsulate my interaction with IO.

The wrapper script:
``` bash
 #! /bin/bash                                                                                                      
                                                                                                                   
 die () {                                                                                                          
     msg="${1:-something has gone horribly wrong}"                                                                 
     printf '%s\n' "$msg" && exit 1                                                                                
 }                                                                                                                 
                                                                                                                   
 for dir in exercises/* ; do                                                                                       
     slug="${dir##*/}"                                                                                             
     slug="${slug//-/_}"                                                                                           
                                                                                                                   
     test_file="${slug}_test.sh"                                                                                   
     test_path="${dir}/${test_file}"                                                                               
                                                                                                                   
     solution="${slug}.sh"                                                                                         
     solution_path="${dir}/${solution}"                                                                            
                                                                                                                   
     config_path="${dir}/.meta/config.json"                                                                        
                                                                                                                   
     [[ -f $solution_path && -f $test_path ]] ||                                                                   
         die "${slug} is missing a file"                                                                           
                                                                                                                   
     cat "$test_path" | bin/configurator "$solution" | jq . > "$config_path"                                       
 done
```

It is pretty lazy I think, but it did the job.

It wrapped around this entry point to `configurator` (of course this leaves out about 90 lines of library I wrote in the `Bats` module):
``` haskell
module Main where                                                                                                 
                                                                                                                   
import System.IO (print, readFile)                                                                                
import System.Environment (getArgs)                                                                               
                                                                                                                   
import Safe (headMay)                                                                                             
                                                                                                                  
import Bats (batsToConfig)                                                                                        
                                                                                                                  
main :: IO ()                                                                                                     
main = (interact . batsToConfig . headMay) =<< getArgs
```

guygastineau:
@bkhl do you have any insights from you key involvement in our progress toward test runner capabilities that would shed light on the right way to deal with the edge cases here? Are they good enough for now?

exercises/ocr-numbers/.meta/config.json
bkhl: > @bkhl do you have any insights from you key involvement in our progress toward test runner capabilities that would shed light on the right way to deal with the edge cases here? Are they good enough for now?

Not sure, it depends a bit on how this ends up being presented in the UI I suppose.

One idea might be to put large input data into separate files, so that these commands (and the tests), can reference those files?

guygastineau: That is something to consider. I have thought about that before, but ultimately it seems like more book keeping on our end.

I guess where I read the link you provided from the research project it seems like both fields in these JSON files are displayed mostly to hell the student locate the test that is failing.

In that case the name is likely sufficient in cases where the command is multiline.

So I suppose I am advocating for simplifying the command string for something like this one.  If that is agreeable I would simply remove the `<<` from the end of these command strings, and clean up the last on from that other file I commented on.

Does such a plan sound like a reasonable way to get things going in the short term @bkhl ?

I will also wait for @glennj 's input.

bkhl: That sounds more like what I _guessed_ the field was for originally, i.e. an identifier or command to run the individual test, not the code run by the test, given the original example:

```
{
  "tests": [
    {
      "name": "MyTestMethod",
      "cmd": "MyTests.MyTestMethod(1)"
    }
  ]
}
```

However, based on https://github.com/exercism/research_experiment_1/issues/32#issuecomment-567836268 it seems the idea is more like that it should be a hint towards running the failing code (although of course running the test would also do that).

My opinion is this should probably be somewhat tailored to what makes most sense for the language. Since with `bats` I don't think there's a way to run just an individual test, we are kind of stuck with showing a way to run the equivalent bash command.

In any case, I think it makes sense to do what you have already done as a starting point, then let's see how it turns out once there is an alpha of the interface to get a clearer picture of how this will be displayed.

guygastineau: Also, somewhat of a side note, I used Haskell for this, because it is comfortable for me.  Also, in the future our best chance of really smart analysis of bash source likely relies on pulling the parser and bash AST module out of the `shellcheck` project (which is all Haskell).

I don't feel like everyone needs to use the work I did for this PR, but we are going to have to figure out how to get better parsing and syntax analysis for v3 to work well in general.

So, obviously I think that the most advanced tools for that job exist in Haskell (even if they need to be harvested from within another project).

I recognize that Haskell remains relatively obscure to most programmers even if it is very powerful.  Therefore I think there are a couple roads for a solution of this AST/analysis front.

1. Python has a mostly complete shell parser.  It is not capable of parsing some more complex situations (I forgot which cases now).  The last time I tried to use this python library I got errors for anything that was not trivial bash.  If some folks who are more into python want to fix this library or extend it to meet our needs, then we could have tools built in a language that is more widely known and installed.
2. As I said before the shellcheck project already has amazing parser and AST modules for bash written in Haskell.  Furthermore, Haskell is an exceptional language for pattern matching and syntax analysis.  I am willing to pull these parts out of shellcheck (they are a bit integrated, so it will take a bit of work to decouple them (especially to get the stack config files working correctly (really the worst part))).  The obvious problem is that requiring a stack installation and knowledge of Haskell for someone to contribute to our analyzers creates a pretty big barrier to entry.  A possible solution:
    - since Haskell is good at language programming we could actually write a EDSL integrated with the bash parsing libs 
    - This should be a simple declarative language that lets a contributor drive the analyzer after learning this simple embedded language instead of having to learn Haskell.
    - The first obvious choice is just to embed a Scheme dialect into it, but I imagine encountering resistance.

In any case, this is likely not the right place to have our this comment, but I can at least copy and paste this to another issuer later if need be.  Regardless, I would like your perspective on this @bkhl as you have been more involved in the analysis side of our repo, and I know you program Python too.

guygastineau:
I think the normalizations in the last commit make sense.  I will check again to see if any other files are affected, but I think these two are the only ones. Overall we have some pretty regular patterns across our 86 test files here ;)

guygastineau:
Another idea that gives less information, but which could help consistency:
    # Just use what they would run in the terminal for the `cmd` part.
    bats ${slug}_test.sh

guygastineau:
@glennj do you have any time to look at this?

glennj:
Now that we have the GHA checks in place, this PR can be closed.

PR #415, opened 2020-01-01
Update bats-core installation via homebrew on OS X

glennj:
Mac *only* change to demonstrate installing bats-core with homebrew. Ubuntu instructions untouched. 

---

Reviewer Resources:

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,9 +1,8 @@
 As there isn't much of a bash ecosystem, there also isn't really a defacto
 leader in the bash testing area. For these examples we are using
-[bats](https://github.com/bats-core/bats-core) (this repo was forked from
-[the original bats](https://github.com/sstephenson/bats) which has
-stagnated).  You should be able to install it from your favorite package
-manager, on OS X with homebrew this would look something like this:
+[bats](https://github.com/bats-core/bats-core). You should be able to
+install it from your favorite package manager, on OS X with
+[Homebrew](https://brew.sh/) this would look something like this:
 
 ```
 $ brew install bats-core
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -1,14 +1,17 @@
-As there isn't much of a bash ecosystem, there also isn't really a defacto leader in the bash testing area. For these examples we are using [bats](https://github.com/sstephenson/bats). You should be able to install it from your favorite package manager, on OS X with homebrew this would look something like this:
+As there isn't much of a bash ecosystem, there also isn't really a defacto
+leader in the bash testing area. For these examples we are using
+[bats](https://github.com/bats-core/bats-core) (this repo was forked from
+[the original bats](https://github.com/sstephenson/bats) which has
+stagnated).  You should be able to install it from your favorite package
+manager, on OS X with homebrew this would look something like this:
 
 ```
-$ brew install bats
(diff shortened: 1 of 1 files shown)


docs/TESTS.md
```diff
@@ -1,14 +1,17 @@
-As there isn't much of a bash ecosystem, there also isn't really a defacto leader in the bash testing area. For these examples we are using [bats](https://github.com/sstephenson/bats). You should be able to install it from your favorite package manager, on OS X with homebrew this would look something like this:
+As there isn't much of a bash ecosystem, there also isn't really a defacto
+leader in the bash testing area. For these examples we are using
+[bats](https://github.com/bats-core/bats-core) (this repo was forked from
+[the original bats](https://github.com/sstephenson/bats) which has
+stagnated).  You should be able to install it from your favorite package
+manager, on OS X with homebrew this would look something like this:
```
guygastineau: Hahaha, so we had the last PR change a couple of lines to be 1 line to match the style you had for the top comment before :rofl: 

Personally I spend most of my time with text using `emacs -nw`, so I definitely prefer multiple lines, but we should give @kotp time to express his opinion :stuck_out_tongue_winking_eye:

guygastineau: BTW, thanks for this PR, @glennj

kotp: I can't keep track of which repositories are "sentence per line" and are "80 columns".  The only thing I really do care about (though I understand the benefit of "sentence per line" and enjoy that as well) is that the style is consistent in the file, at least.  Of course, the greater goal would be repository wide, if not project wide.

kotp: We could remove the history lesson, portion of the change, as this will be something that can be in the repository https://github.com/bats-core/bats-core#background and I don't think we need to keep up on the history if it becomes forked again.

guygastineau:
I am fine with this now.  @kotp ?

docs/TESTS.md
kotp: Looks better, thanks.

docs/TESTS.md
```diff
@@ -1,14 +1,16 @@
...
-==> Downloading from
-https://codeload.github.com/sstephenson/bats/tar.gz/v0.4.0
+$ brew install bats-core
+==> Downloading https://github.com/bats-core/bats-core/archive/v1.1.0.tar.gz
+==> Downloading from https://codeload.github.com/bats-core/bats-core/tar.gz/v1.1.0
 ######################################################################## 100.0%
-==> ./install.sh /opt/boxen/homebrew/Cellar/bats/0.4.0
-/opt/boxen/homebrew/Cellar/bats/0.4.0: 10 files, 60K, built in 2 seconds  
+==> ./install.sh /usr/local/Cellar/bats-core/1.1.0
+🍺  /usr/local/Cellar/bats-core/1.1.0: 13 files, 55KB, built in 4 seconds
```
kotp: Do you know if the "beer" stays there when actually rendered?  Or do we need to escape this so that doesn't happen?

kotp: Oh yeah.  I don't use it that often, but yes, supposed to be there.

glennj:
> Is that a beer I see? Is that meant to be there? If so, then I approve this... Otherwise there may need to be a change here.

That's actually in the brew output.

PR #419, opened 2020-01-17
Update steps for Linux to install bats-core from source

glennj:
Since on Ubuntu, the `bats` package is the old sstephenson one.

---

Reviewer Resources:

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -15,7 +15,18 @@ $ brew install bats-core
 ```
 
 ### For Linux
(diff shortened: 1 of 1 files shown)


guygastineau:
Good catch. I just checked and even `Arch Linux` is using the old repo.

I might see if I can get that to change, but regardless this your changes here are the right ones to make ;)

bkhl:
Pushed two changes:

* Add section on Fedora, which _does_ include a recent version of Bats
* Updated link to point to bats-core

Then noticed that the first thing was fixed in master already, so the force-push was me removing the second change.

PR #424, opened 2020-02-26
fetch_configlet: retry if curl request fails

glennj:
https://exercism-team.slack.com/archives/CAQP7JL3T/p1582614952007300

---

Reviewer Resources:

### DIFF
--- a/bin/fetch_configlet
+++ b/bin/fetch_configlet
@@ -6,7 +6,7 @@ readonly RELEASES='https://github.com/exercism/configlet/releases'
 
 get_version () {
     curl --silent --head ${RELEASES}/latest  |
-        awk -v FS=/ '/Location:/{print $NF}' | tr -d '\r'
+    awk -v FS=/ -v RS='\r\n' 'tolower($1) ~ /location:/ {print $NF}'
 }
 
 case "$(uname)" in
@@ -32,6 +32,7 @@ esac
 
 VERSION="$(get_version)"
 URL="${RELEASES}/download/${VERSION}/configlet-${OS}-${ARCH}.${EXT}"
+localfile=bin/latest-configlet.${EXT}
 
 fetch_archive() {
     local file_info=$1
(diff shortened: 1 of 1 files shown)


bin/fetch_configlet
```diff
@@ -33,11 +33,31 @@ esac
 VERSION="$(get_version)"
 URL="${RELEASES}/download/${VERSION}/configlet-${OS}-${ARCH}.${EXT}"
 
+fetch_archive() {
+    local file_info=$1
+    for i in {0..4}; do
+        sleep $i
```
kotp:
```suggestion
    for delay in {0..4}; do
        sleep $delay
```

IsaacG: ...with variable expansions quoted, of course.

bin/fetch_configlet
```diff
@@ -33,11 +33,31 @@ esac
...
+        curl -s --location "$URL" -o "$localfile"
+        if file "$localfile" 2>&1 | grep -q "$file_info"; then
+            # fetched successfully
```
IsaacG: Shouldn't this just rely on curl's exit status?

glennj: `errexit` is on, so the script will die on non-zero. 
Anyway the problem is that a small text file is fetched, not the (g)zip content we're expecting.

bin/fetch_configlet
glennj: It's an integer, and IFS has not been altered.

glennj:
> Sight reviewed, and approvable, but not tested.

It was (kind of) field tested: https://travis-ci.org/exercism/bash/builds/655446128?utm_source=github_status&utm_medium=notification

glennj:
@kotp @IsaacG @guygastineau I found the fix on another track: while querying the latest version, the `curl --head` results returned the header field names in lower case. I'll keep the retrying code for good measure, but the last commit updates the awk command for case insensitivity.

glennj:
@IsaacG I'm not excluding you, but the UI won't let me re-request your review.

PR #427, opened 2020-03-10
Create exercise: satellite

glennj:
problem spec [canonical-data.json](https://github.com/exercism/problem-specifications/blob/master/exercises/satellite/canonical-data.json)

I didn't port this exercise before because I looked at it and thought "needs OO to build a tree". But after having another look at it, it really just needs to output some JSON.

This is a pretty juicy exercise, as you can see by the number of topics in the config file.

---

Reviewer Resources:


glennj:
@IsaacG I'm not able to add you as a reviewer. I suspect you're not the exercism organization on github.

IsaacG:
Apparently I need to be a track maintainer to be part of the org...

exercises/satellite/README.md
```diff
@@ -0,0 +1,72 @@
+# Satellite
+
+Imagine you need to transmit a binary tree to a satellite approaching Alpha
+Centauri and you have limited bandwidth. Since the tree has no repeating
+items it can be uniquely represented by its [pre-order and in-order traversals][wiki].
```
IsaacG: This isn't how markdown parses links, is it?

exercises/satellite/README.md
```diff
@@ -0,0 +1,72 @@
...
+ / \
+i   x
+   / \
+  f   r
+```
+
+Note: the first item in the pre-order traversal is always the root.
+
+[wiki]: https://en.wikipedia.org/wiki/Tree_traversal
+
```
IsaacG: I think it's worth calling out here what you want th output to look like.
Generating json in bash seems a bit iffy. Might there be a better way to represent it?

exercises/satellite/example.sh
```diff
@@ -0,0 +1,72 @@
...
+    done
+}
+
+indexOf() {
+    local needle=$1
+    local haystack=( "${@:2}" )
+
+    for (( i=0; i < ${#haystack[@]}; i++ )); do
+        if [[ $needle == "${haystack[i]}" ]]; then
+            echo $i
```
IsaacG: Unquoted variable expension. Also, you didn't mark it as local, if you care about that. You can local -i and have a better case against quoting but it's best practice to always quote.

exercises/satellite/example.sh
```diff
@@ -0,0 +1,72 @@
...
+    
+    # read the space-separated items into an array
+    read -ra preorder <<< "$1"
+    read -ra inorder  <<< "$2"
+
+    validate preorder inorder
+
+    if (( ${#preorder[@]} == 0 )); then
+        echo "{}"
+    else
```
IsaacG: You can add a `return` and un-indent the rest of the code here.

exercises/satellite/README.md
kotp: That should be fine.

From Daring Fireball:

```
The title attribute is optional. Link names may contain letters, numbers and spaces, but are not case sensitive:

I start my morning with a cup of coffee and
[The New York Times][NY Times].

[ny times]: http://www.nytimes.com/
Output:

<p>I start my morning with a cup of coffee and
<a href="http://www.nytimes.com/">The New York Times</a>.</p>
```

exercises/satellite/example.sh
glennj: Sure.

exercises/satellite/example.sh
glennj: I get a bit lax with localizing loop variables (I don't use them for any other purpose elsewhere in the code), and I know the content of the variable is an integer so there will be no word splitting, and if the number happens to match a file then echo will still print the same string. I suppose there might be a slight performance benefit if bash doesn't have to attempt split+glob.

Just tested, the -i attribute is no defence against word splitting:
```
$ declare -i i=1234
$ IFS=2
$ printf "%s\n" $i
1
34
```

PR #430, opened 2020-03-21
Consistently use == over = inside [[

IsaacG:
The test files got roughly 2k `[[ ==` and roughly 450 `[[ x =`.
Make that all-or-none :)

---

Reviewer Resources:

### DIFF
--- a/exercises/allergies/allergies_test.sh
+++ b/exercises/allergies/allergies_test.sh
@@ -6,49 +6,49 @@
     #[[ $BATS_RUN_SKIPPED == true  ]] || skip
     run bash allergies.sh 0 allergic_to peanuts
     (( status == 0 )) 
-    [[ $output = "false" ]]
+    [[ $output == "false" ]]
     run bash allergies.sh 0 allergic_to cats
     (( status == 0 )) 
-    [[ $output = "false" ]]
+    [[ $output == "false" ]]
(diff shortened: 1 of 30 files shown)


exercises/allergies/allergies_test.sh
```diff
@@ -6,49 +6,60 @@
     #[[ $BATS_RUN_SKIPPED == true  ]] || skip
     run bash allergies.sh 0 allergic_to peanuts
     [[ $status -eq 0 ]] 
-    [[ $output = "false" ]]
+    [[ $output == "false" ]]
+    [[ $output == "false" ]]
```
glennj: Why are these all doubled?

glennj:
Although `=` is still a valid operator, better messaging to students to use `==`.

exercises/allergies/allergies_test.sh
IsaacG: Fixed

IsaacG:
Personally, I prefer `=` over `==` but I went with what we have more of.

IsaacG:
I might have totally messed up my git history at this point. If I can't figure it out, I may need to delete my fork and just start over.

IsaacG:
Many, many, *many* `git rebase` later, I managed to clean up the history and squash all my commits.

PR #434, opened 2020-03-28
bantor: show line numbers with code.

glennj:
And allow for configurable display.

I trimmed trailing whitespace, so the diff is a little larger than the actual changes.

---

Reviewer Resources:

### DIFF
--- a/mentor_tools/bantor
+++ b/mentor_tools/bantor
@@ -40,7 +40,7 @@ main() {
 
     # don't start logging until we have a workdir
     { date
-      echo "solution: https://exercism.io/mentor/solutions/$uuid" 
+      echo "solution: https://exercism.io/mentor/solutions/$uuid"
     } >"$workdir/bantor.log"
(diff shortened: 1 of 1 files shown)


mentor_tools/bantor
```diff
@@ -167,17 +167,23 @@ I will not do that automatically.
...
     yn "OK to run tests" || return
     BATS_RUN_SKIPPED=true bats "${root}_test.sh"
 }
 
+# Default function to display the code.
+# Can be overridden in the config file if you want a different output.
+displayCode() {
+    cat -n "$1"
+}
+
```
glennj: I use `displayCode() { nl -ba -w4 -s' | ' "$1"; }`

IsaacG: Might be worth considering plumbing in a dotfile/env vars at some point that can be used to control behavior vs overriding function definitions

glennj: I consider that plumbing as already existing.

Issue #438, opened 2020-05-27
Updates on Bats to Bats Core


qaws01395:
I was encountering issues when running bats on my Git bash on Windows. When I read through some conversation online and figured out they have [migrated](https://github.com/sstephenson/bats/pull/234#issuecomment-445023312) to [Bats Core](https://github.com/bats-core/bats-core). 

It works well for me and the testing results looks exactly the same with Bats. Should we update the [Readme](https://github.com/exercism/bash/tree/master/exercises/hello-world#testing) and suggest to use Bats-core instead?

I'm new to open source but I wish to contribute to the community too. Not sure if it's appropriate posting this as a issue.

bkhl:
It seems we started doing that in January but then forgot about it. Started https://github.com/exercism/bash/pull/439 with that branch, with a few additional changes.

PR #439, opened 2020-05-27
Change references for bats to point to bats-core

bkhl:
Looks like we had a forgotten branch from January where we started changing references to Bats to link to https://github.com/bats-core/bats-core

Added another commit to it and changing a few more references, prompted by https://github.com/exercism/bash/issues/438

### DIFF
--- a/docs/TESTS.md
+++ b/docs/TESTS.md
@@ -15,7 +15,18 @@ $ brew install bats-core
 ```
 
 ### For Linux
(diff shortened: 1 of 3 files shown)


kotp:
Are the tests using the new version of Bats?  The changes look good, but might want to confirm it won't break with current tests.

bkhl:
It's what is used for https://github.com/exercism/bash-test-runner , and I've used it when mentoring for some time and had no issues.This merge request got confusing because I accidentally used the same branch name as last time. The first two commits here are actually merged already.

I'll close this and start over.

PR #443, opened 2020-07-02
[TEST CASE] acronym_test.sh

sylenth1cs50:
Test case for apostrophes around a word.

---

Reviewer Resources:

### DIFF
--- a/exercises/acronym/acronym_test.sh
+++ b/exercises/acronym/acronym_test.sh
@@ -65,6 +65,13 @@
   [[ "$output" == "TRNT" ]]
 }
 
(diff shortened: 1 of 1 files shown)


glennj:
Nobody's commented on this yet, so I might as well add my opinion.

I don't think this test case adds much value. The `Halley's comet` one already deals with the apostrophe character (which most solutions I've seen handle by simply removing), and the `The Road __Not__ Taken` one deals with words starting with non-alpha chars.

sylenth1cs50:
Hello Glenn,
Thank you for the revision.
I understand and will close this PR.

PR #453, opened 2020-10-27
Move CI from Travis to GHA

glennj:
Reusing what I did in the Tcl track:

1. GHA workflows for PRs and commits to master in .github/workflows
1. CI scripts in .github/scripts
1. Create bin/validate_one_exercise script containing the guts of bin/validate_exercises -- pull requests only test specific changes.
1. Remove the Travis config

---

Reviewer Resources:


.github/workflows/ci.yml
```diff
@@ -0,0 +1,24 @@
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+
+      - uses: actions/setup-node@v1
+      - name: Install bats-core
+        run: npm install -g bats
+
+      - name: Run tests for all exercises
+        run: bash .github/scripts/ci
```
cmccandless, line 24: If you're calling a script, it's a good idea to run with the following options:
- xtrace (outputs each command that is run, useful for debugging)
- errexit (aborts script with exit code 1 if any command has a non-zero exit code)
	- Does not exit if the failing command is the condition for an `if` or `while`
```suggestion
        run: bash -xe .github/scripts/ci
```

.github/workflows/pr.yml
```diff
@@ -0,0 +1,24 @@
...
+      - uses: actions/setup-node@v1
+      - name: Install bats-core
+        run: npm install -g bats
+
+      - name: Run tests for changed/added exercises
+        run: |
+          PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
+          curl --silent --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
+            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|sh)$")) | .filename' | \
+            xargs -r bash .github/scripts/pr
```
cmccandless, line 24: Same as above.
```suggestion
            xargs -r bash -xe .github/scripts/pr
```

cmccandless (changes requested):

cmccandless:
The checks did not run because they do not yet exist in this repo; you are adding them from a fork. If you have write permissions in this repo, you could open a new branch here and merge your commits from the fork into it, then create a PR from that branch. The checks would then run.

Alternatively, you could merge this, then create a test PR to verify that the checks run as expected. If they are broken, simply PR further fixes.

.github/workflows/ci.yml
glennj: This `ci` script invokes other scripts that already use `-e`, so not necessary to put that option here. 

I get where you're coming from regarding`-x`, but I don't think it would add much value.

glennj:
> The checks did not run because they do not yet exist in this repo;

D'oh, of course. And I can't merge without approval.

cmccandless:
> D'oh, of course. And I can't merge without approval.

I don't currently have write permissions in this repo, so my approval won't be enough.

If @iHiD can approve this PR once the above changes are addressed, I can assist with any issues that may arise afterward.

kotp (approved):

kotp:
I approved early, so the changes can be made, and then brought in without delay.  (Otherwise I can do the `git am` and create the branch, which should also allow the triggers to happen)

bin/validate_one_exercise
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+#
+# For one exercise, 
+# Validate if tests can run properly and the example implementation
+# is passing the tests
```
kotp, line 5: Why does this look so broken?

kotp: (I mean from an "English" standpoint.

glennj: Poor editing

kotp: Cosmetic changes coming shortly.

Issue #479, opened 2021-01-29
merge checks failing due to new exercises directory hierarchy


glennj:
See. https://github.com/exercism/bash/runs/1793817369

The bin/validate_exercises script needs updating to be able to handle the new `practice` subdirectory.

glennj:
Merged.

PR #480, opened 2021-01-29
look for exercises in all subdirectories of exercises/

glennj:
This change will validate all exercises in `exercises/practice/` and in the future `exercises/concept/`  
Addresses issue #479 

---

Reviewer Resources:


iHiD (approved):
Worth nothing that the exercises may differ enough in content that you might need different validators for Concept ones, but this works for now! :)

PR #485, opened 2021-02-12
[v3] Update Markdown files to latest spec

ErikSchierboom:
We've defined a [specification for Markdown files](https://github.com/exercism/docs/blob/main/contributing/standards/markdown.md) to be applied Exercism-wide. This standard includes, amongst others, the following two rules:

- All files must start start with a level-1 heading (`# Some heading text`)
- No heading may decend a level greater than one below the previous (e.g. `## may only be followed by ###, not ####`)

This PR applies the above two rules to the Markdown documents in this repo. 

## Tracking

https://github.com/exercism/v3-launch/issues/17

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,5 @@
+# pull request template
+
 <!-- Your content goes here: -->
 
 
--- a/exercises/practice/gigasecond/.docs/instructions.append.md
+++ b/exercises/practice/gigasecond/.docs/instructions.append.md
@@ -1,3 +1,5 @@
+# Instructions append
+
 ## Test-Driven Development
 
 As programmers mature, they eventually want to test their code.
--- a/exercises/practice/gigasecond/.docs/instructions.append.md
+++ b/exercises/practice/gigasecond/.docs/instructions.append.md
@@ -1,6 +1,4 @@
-# Instructions append
-
-## Test-Driven Development
+# Test-Driven Development
 
 As programmers mature, they eventually want to test their code.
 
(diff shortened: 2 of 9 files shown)


exercises/practice/gigasecond/.docs/instructions.append.md
```diff
@@ -1,3 +1,5 @@
+# Instructions append
+
 ## Test-Driven Development
```
glennj, line 3: @exercism/bash why are we talking about TDD for a non-core exercise that is unlocked about halfway down the core exercises? Surely we can remove this file.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,3 +1,5 @@
+# pull request template
+
 <!-- Your content goes here: -->
```
glennj, line 4: @exercism/bash Do we really need this template for PRs?

exercises/practice/gigasecond/.docs/instructions.append.md
kotp: I believe we talked about this before, and it is to remind students that might have been gone for months or years.  It is something that active users probably become pretty blind to, as they know the routine, but it is there in case they need a reference.

.github/PULL_REQUEST_TEMPLATE.md
kotp: I just made a PR, and this isn't there currently.  I am not sure it is helpful, and gives the submitter no useful information.  It would be prepopulated information and so pretty apparent it came from a template.

I say lines 1- 3 can go away.

Indeed, line 3 doesn't need to go there, commit information ends up somewhere above that line, and though it doesn't appear when submitted, it is It's one of those things that "network traffic doesn't matter as long  as you can't see it" kind of attitudes, which is still traffic that has to be paid for.

It might be helpful to the user if they are new to pull requests to know that they should make some kind of content, but it isn't all that helpful, because if they didn't know what a text area is for, it answers that question, but it doesn't direct to the content that should go there, what "good quality" looks like, etc.  In other words, it is almost a "It makes use feel helpful" line that isn't all that helpful.

glennj: I'll create a separate issue for this.

glennj (approved):

PR #488, opened 2021-02-18
Single Shouted Letter Test for bob

JamesMConroy:
New test case 'A!' that should return "Whoa, chill out!"

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/bob/bob_test.sh
+++ b/exercises/practice/bob/bob_test.sh
@@ -23,6 +23,13 @@
   [[ $output == "Whoa, chill out!" ]]
 }
 
+@test "shouted single letter" {
+  [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+  run bash bob.sh 'A!'
+  (( status == 0 ))
+  [[ $output == "Whoa, chill out!" ]]
+}
+
 @test "asking a question" {
   [[ $BATS_RUN_SKIPPED == "true" ]] || skip
   run bash bob.sh 'Does this cryogenic chamber make me look fat?'


kotp (approved):

kotp:
Approved, but not sure if this catches something where it is commonly missed.

JamesMConroy:
I don't know how common it is, but this case catches a mistake I made while solving the exercise.

So what are the next steps to merge it?

kotp:
Can you explain the mistake made?  Because why not check for 2 characters?  Why not 3? What is special about a single letter, or perhaps that single letter that should be guarded for that isn't covered?

kotp:
The reason why I haven't merged this is because of the following, and it isn't special to Bash:

Any single letter can be its own word in a sentence like this.

We can add it here, but should we also consider moving it up to the site wide specification?

I would like to see the code progression that got to this pull request, to understand if it is a mentoring point, or if it should be removed as a "wiggle room".

JamesMConroy:
Sure. I was checking to see if the input was forceful/all caps using `^[^a-z]{2,}$`.

I that regex was a leftover from a stratedgy where I was trying to create an exclusive regex for each case.

When the mentor pointed out that it would miss the input 'A!', I added the test case and implimented a much simplier regex `^[A-Z]+$`.

kotp:
Thanks.  So would you say you would have picked up on that if the test had been there? Would it have hidden a learning moment?  So should we really add the test?

Maintainer/Mentor input welcome as well.

kotp (approved):

IsaacG:
@JamesMConroy Are you still working on this PR or should it be marked abandoned?

kotp:
Abandoned is really something that I would have not expected the OP to respond to.

I have labelled it "wontfix" because I am assuming the conflicts will not be attended to, not because it is not a potential solution to a potential general issue with the exercise.

PR #491, opened 2021-02-28
Change brew install command to bats-core

coreygo:
The linked repository and command output mismatched the command shown which might lead to confusion if users miss the bash tests documentation which also points to the correct bats-core install.

---

Reviewer Resources:


glennj (approved):
Good catch!

PR #505, opened 2021-05-05
Add CODEOWNER for fetch configlet script

ErikSchierboom:
This PR adds the fetch configlet script(s) to the `.github/CODEOWNERS` file.

The fetch configlet script(s) are copied of the files in the https://github.com/exercism/configlet/blob/main/scripts directory.
We want to prevent tracks from changing these fetch scripts as they should be updated centrally, to have updates benefit all tracks.

## Tracking

https://github.com/exercism/configlet/issues/286


.github/CODEOWNERS
```diff
@@ -3,3 +3,6 @@ config/maintainers.json @exercism/maintainers-admin
 
 # Code owners
 .github/CODEOWNERS      @exercism/maintainers-admin
+
+# Changes to `fetch_configlet` should be made in the `exercism/configlet` repo
```
ee7, line 7: This line is a bit weird currently, given that the files are quite different.

ee7 (commented):
Note that the bash track's [`fetch_configlet`](https://github.com/exercism/bash/blob/51cd42984a4f337125dd1c5cc4e4c4f592bf3982/bin/fetch_configlet) is different from the upstream [`fetch-configlet`](https://github.com/exercism/configlet/blob/792bdade34f450e1ccc3f66be840bead03fa0faa/scripts/fetch-configlet).

In the long-term I'd suggest that the two files become identical, which might involve integrating any improvements from the bash track's version into upstream.

ErikSchierboom:
> In the long-term I'd suggest that the two files become identical, which might involve integrating any improvements from the bash track's version into upstream. (I haven't looked at the bash track's version right now).

A quick look suggests that the only real change was some delay to deal with files not being downloaded. I'm fairly sure those were caused by rate limits, which the actual `fetch-configlet` script fixes, so I don't think we need to integrate any changes upstream

I _would_ like to have this file be named the same across repos, as then we can auto-sync it across all tracks. @glennj is there anything fundamentally against renaming the `fetch_configlet` to `fetch-configlet` and to update its contents to the latest version? If you'd really like the also have the `fetch_configlet` script, maybe a symlink could be used?

glennj:
I have no lingering love for the previous script. I'd be happy to have it deleted, and we have one fetch to rule them all.

ErikSchierboom:
@glennj I've added commit https://github.com/exercism/bash/pull/505/commits/7c4533b51613c3695efae56447dc09860be6900a to the PR to replace the old script with the new one.

ee7 (approved):
> A quick look suggests that the only real change was some delay to deal with files not being downloaded. I'm fairly sure those were caused by rate limits, which the actual `fetch-configlet` script fixes, so I don't think we need to integrate any changes upstream

Also, the upstream `fetch-configlet` script calls `curl` with `--retry 3` anyway.

---

LGTM. I don't care whether this PR is squashed or not - just note that if you squash,
```
Add CODEOWNER for fetch configlet script
```
is now a bad commit message.

And it turns out that this PR also fixes the `CONTRIBUTING.md` instructions:

https://github.com/exercism/bash/blob/51cd42984a4f337125dd1c5cc4e4c4f592bf3982/CONTRIBUTING.md#L220-L222

glennj (approved):
I like it.

glennj:
@iHiD this PR is stuck awaiting approval. Can you have a look?

iHiD (approved):

PR #530, opened 2021-08-17
trigger a rebuild because linter failed on old pr.

MichaelDimmitt:
the readme on the hello world exercise is recommending installing 
$ brew install bats

when it should be the supported package.
$ brew install bats-core

https://github.com/exercism/bash/pull/491/checks
---

Reviewer Resources:

### DIFF
--- a/exercises/practice/hello-world/.docs/instructions.append.md
+++ b/exercises/practice/hello-world/.docs/instructions.append.md
@@ -25,7 +25,7 @@ install it from your favorite package manager, on OS X with homebrew
 this would look something like this:
 
 ```
-$ brew install bats-core
+$ brew install bats-core 
 ==> Downloading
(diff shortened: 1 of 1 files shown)


SleeplessByte:
They pass. Assuming you don't actually want to merge this? (Re-open if you do, but that seems weird with the change as provided).

MichaelDimmitt:
How can we confirm that the hello world exercise has the correct text showing ?

@SleeplessByte 
When I look under my excercises it shows `brew install bats` when it should say `brew install bats-core` according to this document unless the pr should exist somewhere else?

Here is a community solution posted today august 17 for reference.
I realize this probably should have been an issue. I was just hoping triggering a rebuild would cause a deploy to occur.

MichaelDimmitt:
No need to re-open the pr,
should I open an issue about this ?

PR #539, opened 2021-09-03
add bats-assert functionality

### LINKED ISSUE #531: Add diffing support to bash tests
Diffing is easy to attain in bats
If you add printDiff to every function. 
It will only show the output if the test fails.

curious if @IsaacG has thoughts.

by changing
```bash
@test "a name given" {
  [[ $BATS_RUN_SKIPPED == true  ]] || skip
  run two_fer Alice
  [[ $status -eq 0 ]]
  [[ $output == "One for Alice, one for me." ]]
}
```
```bash
printDiff(){
  echo output;
  echo "$output";
  echo ;
  echo expected;
  echo "$expected;
}

@test "a name given" {
  [[ $BATS_RUN_SKIPPED == true  ]] || skip
  run two_fer Alice

  expected=$(cat <<-END
One for Alice, one for me.
END
  )

  printDiff
  [[ $status -eq 0 ]]
  [[ "$output" ==  "$expected" ]]
}
```

This actually just saved me on a project.
And I am sourcing the scripts like libs as is proposed in the pending pr
https://github.com/exercism/bash/pull/373

### LINKED ISSUE #536: add bats-assert and bats-file plugins to bats test suite
I created a request in bats-core to add homebrew support for their plugins using homebrew taps.

https://github.com/bats-core/bats-core/issues/492

In the meantime, 
https://github.com/bats-core/bats-assert

Refers to 
https://github.com/ztombol/bats-docs

Which uses the homebrew tap from kaos
https://github.com/kaos/homebrew-shell

```sh
$ brew tap kaos/shell
```

```sh
$ brew install bats-assert
$ brew install bats-file
```

then in every test you have to load the files.
```bash
load '/opt/homebrew/lib/bats-support/load.bash'
load '/opt/homebrew/lib/bats-assert/load.bash'

@test 'refute_output()' {
  run echo 'want'
  refute_output 'want'
}
@test 'assert_equal()' {
  assert_equal 'have' 'hav'
}
```
Screen Shot 2021-08-26 at 8 57 52 PM

glennj:
Resolves #536. Resolves #531 

The `bats-extra.bash` will be added to every exercise. The contents are the `src` files for bats-support and bats-assert concatenated, with comment blocks removed.

Each exercise's test file will be updated to use bats-assert functions.

The augmented test failure output is important particularly for the 

---

Reviewer Resources:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -17,6 +17,8 @@ jobs:
         run: sudo npm install -g bats
 
       - name: Run tests for changed/added exercises
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
           pr_endpoint=$(jq -r '"repos/\(.repository.full_name)/pulls/\(.pull_request.number)"' "$GITHUB_EVENT_PATH")
           gh api "$pr_endpoint/files" --paginate --jq '
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -24,6 +24,6 @@ jobs:
           gh api "$pr_endpoint/files" --paginate --jq '
             .[] | 
               select(.status == "added" or .status == "modified") |
-              select(.filename | match("\\.(md|sh)$")) |
+              select(.filename | match("\\.(md|sh|bash)$")) |
               .filename
           ' | xargs -r bash .github/scripts/pr
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -18,7 +18,10 @@ jobs:
 
       - name: Run tests for changed/added exercises
         run: |
-          PULL_REQUEST_URL=$(jq -r ".pull_request.url" "$GITHUB_EVENT_PATH")
-          curl --silent --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|sh)$")) | .filename' | \
-            xargs -r bash .github/scripts/pr
+          pr_endpoint=$(jq -r '"repos/\(.repository.full_name)/pulls/\(.pull_request.number)"' "$GITHUB_EVENT_PATH")
+          gh api "$pr_endpoint/files" --paginate --jq '
+            .[] | 
+              select(.status == "added" or .status == "modified") |
+              select(.filename | match("\\.(md|sh)$")) |
+              .filename
+          ' | xargs -r bash .github/scripts/pr
--- a/exercises/practice/hello-world/bats-extra.bash
+++ b/exercises/practice/hello-world/bats-extra.bash
@@ -1,0 +1,637 @@
+# This is the source code for bats-support and bats-assert, concatenated
+# * https://github.com/bats-core/bats-support
+# * https://github.com/bats-core/bats-assert
+#
+# Comments have been removed to save space. See the git repos for full source code.
+
+############################################################
+#
+# bats-support - Supporting library for Bats test helpers
+#
+# Written in 2016 by Zoltan Tombol <zoltan dot tombol at gmail dot com>
+#
+# To the extent possible under law, the author(s) have dedicated all
+# copyright and related and neighboring rights to this software to the
+# public domain worldwide. This software is distributed without any
+# warranty.
+#
+# You should have received a copy of the CC0 Public Domain Dedication
+# along with this software. If not, see
+# <http://creativecommons.org/publicdomain/zero/1.0/>.
+#
+
+fail() {
+  (( $# == 0 )) && batslib_err || batslib_err "$@"
+  return 1
+}
+
+batslib_is_caller() {
+  local -i is_mode_direct=1
+
+  # Handle options.
+  while (( $# > 0 )); do
+    case "$1" in
+      -i|--indirect) is_mode_direct=0; shift ;;
+      --) shift; break ;;
+      *) break ;;
+    esac
+  done
+
+  # Arguments.
+  local -r func="$1"
+
+  # Check call stack.
+  if (( is_mode_direct )); then
+    [[ $func == "${FUNCNAME[2]}" ]] && return 0
+  else
+    local -i depth
+    for (( depth=2; depth<${#FUNCNAME[@]}; ++depth )); do
+      [[ $func == "${FUNCNAME[$depth]}" ]] && return 0
+    done
+  fi
+
+  return 1
+}
+
+batslib_err() {
+  { if (( $# > 0 )); then
+      echo "$@"
+    else
+      cat -
+    fi
+  } >&2
+}
+
+batslib_count_lines() {
+  local -i n_lines=0
+  local line
+  while IFS='' read -r line || [[ -n $line ]]; do
+    (( ++n_lines ))
+  done < <(printf '%s' "$1")
+  echo "$n_lines"
+}
+
+batslib_is_single_line() {
+  for string in "$@"; do
+    (( $(batslib_count_lines "$string") > 1 )) && return 1
+  done
+  return 0
+}
+
+batslib_get_max_single_line_key_width() {
+  local -i max_len=-1
+  while (( $# != 0 )); do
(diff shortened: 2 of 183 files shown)


glennj:
For example, "hello world":

The successful test output appearance does not change
```bash
$ bats hello_world_test.sh
 ✓ Say Hi!

1 test, 0 failures
```
If the exit status is non-zero:
```bash
$ cat hello_world.sh
#!/usr/bin/env bash

echo "Hello, World!"
exit 2

$ bats hello_world_test.sh
 ✗ Say Hi!
   (from function `assert_success' in file ./bats-extra.bash, line 409,
    in test file hello_world_test.sh, line 10)
     `assert_success' failed with status 2

   -- command failed --
   status : 2
   output : Hello, World!
   --


1 test, 1 failure

$ echo $?
1
``` 
If the output does not match what's expected:
```bash
$ cat hello_world.sh
#!/usr/bin/env bash

echo "Goodbye, Mars!"

$ bats hello_world_test.sh
 ✗ Say Hi!
   (from function `assert_output' in file ./bats-extra.bash, line 394,
    in test file hello_world_test.sh, line 13)
     `assert_output "Hello, World!"' failed

   -- output differs --
   expected : Hello, World!
   actual   : Goodbye, Mars!
   --


1 test, 1 failure

$ echo $?
1
```

glennj:
@exercism/bash Thoughts please?

exercises/practice/hello-world/bats-extra.bash
```diff
@@ -0,0 +1,637 @@
...
+
+batslib_print_kv_single() {
+  local -ir col_width="$1"; shift
+  while (( $# != 0 )); do
+    printf '%-*s : %s\n' "$col_width" "$1" "$2"
+    shift 2
+  done
+}
+
+batslib_print_kv_multi() {
```
kotp, line 99: What is `kv` in the name?  And `batslib` or should it be two words, as `bats_lib` for consistency?  It appears most words are separated by underscores throughout.

If `kv` represents "key value" or "key and value" what does the key represent and what does the value represent?

kotp (approved):
Looks good to me, with one suggestion for the repeated "kv" use in the names to show what the placeholder might be.

exercises/practice/hello-world/hello_world_test.sh
```diff
@@ -1,10 +1,14 @@
 #!/usr/bin/env bash
+source ./bats-extra.bash
```
IsaacG, line 2: Any test to fail if the file is missing or cannot be sourced?

exercises/practice/hello-world/bats-extra.bash
```diff
@@ -0,0 +1,637 @@
+# This is the source code for bats-support and bats-assert, concatenated
+# * https://github.com/bats-core/bats-support
+# * https://github.com/bats-core/bats-assert
+#
+# Comments have been removed to save space. See the git repos for full source code.
```
IsaacG, line 5: Is this file entirely upstream code (i.e. I shouldn't bother reviewing) or is it only partial upsteam?

IsaacG (changes requested):

exercises/practice/hello-world/bats-extra.bash
glennj: I didn't write any of that, and I don't indent to change anything in it.

That particular function is at https://github.com/bats-core/bats-support/blob/master/src/output.bash#L136

exercises/practice/hello-world/bats-extra.bash
glennj: It's all of https://github.com/bats-core/bats-support/tree/master/src and https://github.com/bats-core/bats-assert/tree/master/src

exercises/practice/hello-world/hello_world_test.sh
glennj: Hmm, well if it can't be found then `exercism download` would be broken. If the student accidentally deletes it, then a helpful message to re-download would be useful

glennj: I changed `source` to the bats function `load`. If the file does not exist, then
```bash
$ bats hello_world_test.sh
 ✗ File hello_world_test.sh
bats: bats-extra.bash does not exist
   (in test file hello_world_test.sh, line 9)
     `load bats-extra.bash' failed

1 test, 1 failure
```

IsaacG (approved):

kotp (approved):

kotp:
I can see that it is titled WIP, but "convert to draft" is probably a better solution.  It can prevent premature merges.

glennj:
Phew, all tests passing.

glennj:
> I can see that it is titled WIP, but "convert to draft" is probably a better solution. It can prevent premature merges.

No longer WIP/draft

glennj:
Reviewers: coments about the commits here:

commit | msg | notes
-|-|-
dd41d86 | WIP: add bats-assert functionality | convert "hello world"
646c5fd | use bats "load" function instead of "source" | errors out if bats-extra.bash not source-able
05954bd | add bats-extra.bash to all exercises | adding file to all other exercises
ebd194c | transform test scripts to use bats-assert functions | this is big delta and where you'll want to spend your time: convert all test suites to use assert functions
5ae639f | with an associative array, "${map[@]}" order is indeterinate, broke i… | fix to example script for bash 5.1
f2fc14c | in arithmetic, bare map entry with quote in key broke in bash 5.1| fix to example script for bash 5.1
54d7be2 | update workflow, using github cli with auto-pagination, to pull in AL… | update GHA workflow script to work with loads of changes
2495798 | add env var to workflow for auth | update GHA workflow script to work with loads of changes

kotp (approved):

.github/workflows/pr.yml
```diff
@@ -17,8 +17,13 @@ jobs:
...
-          curl --silent --url $"${PULL_REQUEST_URL}/files" --header 'authorization: Bearer ${{ secrets.GITHUB_TOKEN }}' | \
-            jq -c '.[] | select(.status == "added" or .status == "modified") | select(.filename | match("\\.(md|sh)$")) | .filename' | \
-            xargs -r bash .github/scripts/pr
+          pr_endpoint=$(jq -r '"repos/\(.repository.full_name)/pulls/\(.pull_request.number)"' "$GITHUB_EVENT_PATH")
+          gh api "$pr_endpoint/files" --paginate --jq '
+            .[] | 
+              select(.status == "added" or .status == "modified") |
+              select(.filename | match("\\.(md|sh)$")) |
+              .filename
+          ' | xargs -r bash .github/scripts/pr
```
IsaacG, line 29: You can lose the xargs and just do this in a loop with bash.

```
mapfile -t filenames < <(gh api ...)
for filename in "${filenames[@]}"; do
  bash ... "$filename"
done
```
```

**Review by IsaacG — APPROVED** [review]

File: .github/workflows/pr.yml (hunk 4 above)
Comment (SleeplessByte) [reply .github/workflows/pr.yml:29]: We _should_ be able to use `git diff` directly:

```shell
git diff --name-only
```

glennj: @IsaacG I could, but what's the advantage over xargs? Do we know that the shell used to execute the `run` item is bash so that mapfile is even available?

IsaacG: Do we know that the machine has xargs installed? 

`cmd | while read filename ...` is the POSIX-sh version of that if you want to support `sh`. Though at that point, should `xargs` be calling `bash` or `sh`?

glennj: > Do we know that the machine has xargs installed? 

Yes: https://github.com/exercism/bash/pull/539/checks?check_run_id=3514619188#step:5:1

MichaelDimmitt:
cool, thanks for adding this. 

nice idea to save users from needing to install the posix compliant bash in different environments.
since it is posix compliant we can just add the files for them.

and adding the extra file bats-extra.bash that
concatenated bats assert and bats support

In the future maybe a script iterates through all files in the test folder and updates bats-extra.bash if the plugins have an update we want to add. But no need to overengineer at present moment.

definately gives more output to students 👍

glennj:
Thanks for the inspiration. While I was working on it I was thinking I should apologize: I think I was pretty dismissive about your ideas. So, sorry. Working with the v3 website really revealed the need for better test failure feedback.

PR #541, opened 2021-09-15
test files: rename xxx_test.sh to xxx.bats, fix shebang

braoult:
tests scripts: change extension to .bats, fix shebang.

---

Reviewer Resources:


iHiD (dismissed):
**This shouldn't get merged until https://github.com/exercism/bash-test-runner/pull/27 is merged**

We need to be careful on this. The test runner most support both existing exercises and this new format, and that support must be deployed, before this is merged.

I'm "Requesting Changes" just to block this until the test runner is merged.

iHiD:
@angelikaTyborska, @neenjaw, @mPizenberg Any advice on this after doing it on your tracks?

braoult:
> We need to be careful on this. The test runner most support both existing exercises and this new format, and that support must be deployed, before this is merged.

https://github.com/exercism/bash-test-runner/pull/27 supports both old (`xxx_test.sh`) and new (`xxx.bats`) file names. And yes, this PR was planned to be hold until bash-test-runner PR is merged.

mpizenberg:
tbh I haven't fully taken the time yet to evaluate the implications of the change on the online tooling. The tooling itself in the elm track does not support the old way, only the new one, and the filepaths changes actually occurred a while ago, before starting the work on v3. It's just that I noticed the multiple files issues on the online editor for even older submissions of mine.

braoult:
> tbh I haven't fully taken the time yet to evaluate the implications of the change on the online tooling. The tooling itself in the elm track does not support the old way, only the new one, and the filepaths changes actually occurred a while ago, before starting the work on v3. It's just that I noticed the multiple files issues on the online editor for even older submissions of mine.

When I discussed about this change on Slack yesterday, I wondered what could be some other side effects, nobody raised the online tooling... Could this be a blocking point ? I don't know at all  how all of this interacts...

mpizenberg:
Well, the online tooling is composed of the tests runner (seems you took care of that), the analyzer (it will have to take into account both in the future if it's going to be run on old solutions), the representer (which only cares about the solution, not the tests so fine here) and the online editor, which needs to present the solution (and tests for practice exercises) to the user.

So in you case, I think the only point left is to check how is handled the presentation of the tests file in the online editor for practice exercise.

braoult:
> So in you case, I think the only point left is to check how is handled the presentation of the tests file in the online editor for practice exercise.

Yes.

@iHiD, why the `bash` (student' solution) files would be read-only, and not the `bats`files (tests) ? I am not sure to understand. What we do here is:
- before :
```
student solution: xxx.sh
test program: xxx_test.sh
```
- after :
```
student solution: xxx.sh
test program: test.bats
```

angelikatyborska:
> @angelikatyborska Any advice on this after doing it on your tracks?

I actually started maintaining the track _after_ this change happened 🤷

braoult:
My concern here is only : _Where_ and _when_ the online editor gets the list of files to edit from ?
- *where*: Is there some hardcoded names on UI side, or is the list taken from somewhere (likely `.meta/config.json` ?). If it is taken from `config.json`, I don't see any issue (I updated them in this PR). If not, I simply have no clue on what should be done there.
- *when* : I hope the filenames information is part of the student's solution environment (i.e. the `filenames` are determined when the student starts to use UI, and never changed later). Is that the way it works ?

angelikatyborska:
> where: Is there some hardcoded names on UI side, or is the list taken from somewhere (likely .meta/config.json ?).

It is taken from `.meta/config.json`, the three types of files - "test", "solution", and "editor" (non-editable files). But added to that are the files that the student submitted for the previous solution. Via the web editor they only could have submitted the previous values of `files.solution`, but via the CLI anything.

> when : I hope the filenames information is part of the student's solution environment (i.e. the filenames are determined when the student starts to use UI

@iHiD will know this best, but my understanding is that every solution contains the git SHA of the exercise as it were when the student started the exercise and takes files directly from git from that point in time. However, it can change. The platform shows when a change is done to an exercise and shows you the diff between your version and the new version. Then, you can choose to update your solution to the newest version and I think that changes the SHA which means the student will now get the newest versions of all files, including the new `.meta/config.json`.

iHiD:
OK, a couple of things.

1. The files are locked to what the student has when they start the exercise, but they are prompted to update when changes are made, at which point they get multiple solution files if the solution file has been renamed. That's because they get all their submitted files, and all the files in the `solution` key.
1. I misunderstood and thought that we were renaming the stubs, not the tests. Sorry. The fact that only the tests are changing makes this less of a concern. I'm not sure what happens when a student updates to the latest version then downloads that update. I doubt it will delete their old tests file, so they'll have both. Will that be ok?

I've now merged the test-runner one so remove my lock on this.

braoult:
> I'm not sure what happens when a student updates to the latest version then downloads that update. I doubt it will delete their old tests file, so they'll have both. Will that be ok?

Yes: If both files are there, test-runner will use by default the new one (`xxx.bats`). This is done in `bin/run.sh` :
```
    local test_file="${slug//-/_}.bats"
    # test scripts may be (old) xxxx_test.sh
    [[ -f "$test_file" ]] || test_file="${slug//-/_}_test.sh"
```
As the test code itself is unchanged, we are fine here.

glennj (changes requested):

exercises/shared/.docs/tests.md
```diff
@@ -3,7 +3,7 @@
 Each exercise contains a test file.
 Run the tests using the `bats` program.
 ```bash
-bats hello_world_test.sh
+bats hello_world.bats
 ```
 
 `bats` will need to be installed.
```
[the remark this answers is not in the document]
braoult: Ooops. Fixed.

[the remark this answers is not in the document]
braoult: Oh, I also removed ending spaces... Is it OK ?

glennj (approved):

glennj:
The workflow is doing this:
```bash
gh api repos/exercism/bash/pulls/541/files --paginate --jq '
  .[] |
    select(.status == "added" or .status == "modified" or .status == "renamed") |
    select(.filename |
    match("\\.(sh|bash|bats)$")) |
    .filename
'
```

glennj:
@IsaacG I know you have opinions about this subject. Can you review please?

braoult:
> @IsaacG I know you have opinions about this subject. Can you review please?

Do you mean the extension itself, as seen on Slack (no extension for executable) ?

glennj:
> Do you mean the extension itself, as seen on Slack (no extension for executable) ?

No, removing the extension will make the pr.yml github workflow harder to make specific.

bin/validate_one_exercise
```diff
@@ -32,4 +32,4 @@ trap cleanup EXIT
 cp "${solution}" "${exercise_name}.sh"
 
 # Run the tests
-bats "${exercise_name}_test.sh"
+bats "${exercise_name}.bats"
```
IsaacG, line 35: Side note/no change requested. In theory, with the fixed shebang. we should be able to give the tests a `+x` and just run them ... but in practice calling `bats` with the filename is probably easier.

IsaacG (approved):
This looks good to me! Thank you for your work.

bin/validate_one_exercise
braoult: Exact. I thought about it, but unsure about this `+x` on non *ix-like systems...

glennj: Also, I don't think that exercism download respects the executable bits.

PR #542, opened 2021-09-23
Change assert output for empty strand cases

hasithsen:
Messages changed to fit empty strand cases.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/hamming/hamming.bats
+++ b/exercises/practice/hamming/hamming.bats
@@ -58,14 +58,14 @@ load bats-extra
   [[ $BATS_RUN_SKIPPED == "true" ]] || skip
   run bash hamming.sh '' 'G'
   assert_failure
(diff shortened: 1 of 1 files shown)


glennj:
Is there a reason for this? This goes against the central problem specifications: https://github.com/exercism/problem-specifications/blob/main/exercises/hamming/canonical-data.json

glennj:
Changes to the test scripts must also be accompanied by the corresponding changes to the `.meta/example.sh` file. In your fork, at the root directory of the track repo, run 
```bash
bin/validate_one_exercise exercises/practice/hamming
```
to duplicate what the github checks do.

hasithsen:
@glennj thanks for the info! We currently allow empty strands when both left and right strings are empty. But we expect the output to fail when one strand is empty and the other is not—strands must be of equal length. Therefore, closing since this change isn't necessary, thanks again for the info!

PR #552, opened 2021-10-06
add new Markdown test cases

glennj:
Checks headers for h3, h4, h5, and validates that 7 hashes don't generate a h7

Recall that this is a refactoring exercise, so the stub file must pass the test suite.  
Fix was required for example solution too.


kotp (approved):
:shipit:

PR #559, opened 2021-12-12
add explicit test command?

Kidlike:
I wasn't sure if this is a good idea or not, as other tracks don't have this. But again, other tracks have more established testing concepts.

Since this is a "hello world" exercise, it could be discouraging that there are no instructions.

I hope this has not been discussed before, but I couldn't find anything except people being very confused about what they are doing wrong. :man_shrugging: 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/hello-world/.docs/instructions.md
+++ b/exercises/practice/hello-world/.docs/instructions.md
@@ -9,7 +9,7 @@ or environment.
 The objectives are simple:
 
 - Write a function that returns the string "Hello, World!".
-- Run the test suite and make sure that it succeeds: `bats hello_world.sh`
+- Run the test suite and make sure that it succeeds: `bats hello_world.bats`
 - Submit your solution and check it at the website.
 
 If everything goes well, you will be ready to fetch your first real exercise.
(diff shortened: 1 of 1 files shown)


exercises/practice/hello-world/.docs/instructions.md
```diff
@@ -9,7 +9,7 @@ or environment.
 The objectives are simple:
 
 - Write a function that returns the string "Hello, World!".
-- Run the test suite and make sure that it succeeds.
+- Run the test suite and make sure that it succeeds: `bats hello_world.sh`
```
glennj, line 12: Needs to be `bats hello_world.bats`

glennj: Perhaps also link to the testing documentation:
```
[Run the test suite][testing] and make sure ...

[testing]: https://exercism.org/docs/tracks/bash/tests
```

glennj (changes requested):

exercises/practice/hello-world/.docs/instructions.md
Kidlike: well, that was an embarrassing typo :)

adjusted :+1:

kotp (approved):
The introductory exercise does need to be a little more self contained, so I think this is good.

I like that it also links to the documentation page as well.

glennj (approved):

glennj:
Thank you very much @Kidlike

PR #560, opened 2021-12-24
Update test cases

glennj:
Pull in all latest changes from problem-specs


kotp (approved):
Looks good.  The last commit took the longest to review, but I did not see any changes that did not make sense and seems to improve what is there.

glennj:
The "sync docs" and "sync metadata" commits come directly from changes in problem specs

PR #562, opened 2021-12-26
add "Pure Bash Bible" link to the list of "Useful Bash resources"

meleu:
In the bash track the use of pure-bash solutions is encouraged (rather then using external processes like `cut`, `grep`, `sed`, `awk`, etc.). For this purpose I consider the [Pure Bash Bible](https://github.com/dylanaraps/pure-bash-bible) an invaluable resource.

---

Reviewer Resources:

### DIFF
--- a/docs/RESOURCES.md
+++ b/docs/RESOURCES.md
@@ -4,6 +4,7 @@
 - [Bash Guide](http://mywiki.wooledge.org/BashGuide)
 - [Bash FAQ](http://mywiki.wooledge.org/BashFAQ)
 - [Bash Pitfalls](https://mywiki.wooledge.org/BashPitfalls)
+- [Pure Bash Bible](https://github.com/dylanaraps/pure-bash-bible) - collection of pure bash alternatives to external processes.
 - [GNU Bash Manual](http://gnu.org/s/bash/manual)
 - [Bash-Hackers Wiki](http://wiki.bash-hackers.org)
 - [Shellcheck provides shell script guidelines and help](http://www.shellcheck.net/)


docs/RESOURCES.md
```diff
@@ -4,6 +4,7 @@
 - [Bash Guide](http://mywiki.wooledge.org/BashGuide)
 - [Bash FAQ](http://mywiki.wooledge.org/BashFAQ)
 - [Bash Pitfalls](https://mywiki.wooledge.org/BashPitfalls)
+- [Pure Bash Bible](https://github.com/dylanaraps/pure-bash-bible) - collection of pure bash alternatives to external processes.
```
kotp, line 7: One of these things are not like the others.  This is the only one with an additional comment for the bullet list.

I am not saying that it should not be here, but wonder if it should be here.

kotp (approved):
Approving because the additional information is not a deal breaker.

docs/RESOURCES.md
kotp: See line 9 where the additional information is in the link description, rather than as an additional text after the link.

IsaacG:
The very first example needlessly modifies `$_` which ... isn't great. `$_` has a specific meaning and use. If they want a local variable, they should use it. I find the nested expansions hard to read. If you want to trim spaces, `read` works great for that.

The second example reaches for `set -f` and `set --` to trim whitespace.  `read -r var <<< "$*"` works just fine to trim whitespace.

There's a *ton* of bash resources out there. The number of high quality resources, on the other hand, is much, much lower. I don't think resources should be added just because they exist. 

https://wiki.bash-hackers.org/scripting/tutoriallist is a pretty decent list of community-vetted and community-ranked resources. It might make sense to just link to there and not try to do our own ranking.

kotp:
I vote for the bash hackers tutorial list, as well.  That said, there are, though during mentoring, links that I do give that have information that I do not fully agree with for the benefit of contrast and the useful information that is there that I do agree with.

This document does not have that benefit of guidance, and so it should be something we should hesitate to add to.

IsaacG:
Sounds like we're not in favor of this particular PR. Will close it out.
Thank you for your suggestion and participation!

meleu:
> The very first example needlessly modifies `$_` which ... isn't great. `$_` has a specific meaning and use. If they want a local variable, they should use it. I find the nested expansions hard to read. If you want to trim spaces, `read` works great for that.

Agreed.


> The second example reaches for `set -f` and `set --` to trim whitespace. `read -r var <<< "$*"` works just fine to trim whitespace.

Your solution doesn't solve the problem to be solved in the second example. Which is to trim out leading/trailing white space **and** truncate spaces in the middle of the string.


> There's a _ton_ of bash resources out there. The number of high quality resources, on the other hand, is much, much lower. I don't think resources should be added just because they exist.

I agree that resources shouldn't be added just because they exist.

I suggested this specific one because I think it's a valuable bash resource. It fits specially well in the exercism's list of "Useful Bash resources" because [using pure-bash solutions are encouraged here](https://github.com/exercism/bash/blob/main/exercises/shared/.docs/help.md#external-utilities).


> https://wiki.bash-hackers.org/scripting/tutoriallist is a pretty decent list of community-vetted and community-ranked resources. It might make sense to just link to there and not try to do our own ranking.

Your suggestion here seems to fit well in the "[How to learn Bash](https://exercism.org/docs/tracks/bash/learning)" doc.


> Sounds like we're not in favor of this particular PR.

According to the discussions above one person approved the PR and another one was against it.

I'm not aware about how things are decided around here, but if there other bash maintainers, it would be interesting to see their input too. If it's possible, I'd like to respectfully ask to reopen this PR.

As I said, the "Pure Bash Bible" seems to be very well aligned with [what we encourage here](https://github.com/exercism/bash/blob/main/exercises/shared/.docs/help.md#external-utilities).

glennj:
It's got some interesting stuff in it. (IMO, hubris to call it a "bible", but that's apropos of nothing). I'd suggest to add it as a separate paragraph after the list. Something like:

> Since the goal of this track is to focus on pure-bash solutions to the exercises, one handy resource is the [Pure Bash Bible](https://github.com/dylanaraps/pure-bash-bible), showing bash-only methods to solve some common problems.

IsaacG:
> > The second example reaches for `set -f` and `set --` to trim whitespace. `read -r var <<< "$*"` works just fine to trim whitespace.
> 
> Your solution doesn't solve the problem to be solved in the second example. Which is to trim out leading/trailing white space **and** truncate spaces in the middle of the string.

My point being, there's better approaches for a lot of these than the way this "bible" does it. At a glance, it doesn't look very high quality and I'm reluctant to point students at subpar syntax and solutions as recommended learning material.

> I suggested this specific one because I think it's a valuable bash resource. It fits specially well in the exercism's list of "Useful Bash resources" because [using pure-bash solutions is encouraged here](https://github.com/exercism/bash/blob/main/exercises/shared/.docs/help.md#external-utilities).

There's other resources that are higher quality. There's a *lot* of resources out there and this isn't an exemplar resource.


> According to the discussions above one person approved the PR and another one was against it.

One person was against it. The other person approved ... then wrote,

> This document does not have that benefit of guidance, and so it should be something we should hesitate to add to.

That's two people hesitant to add it and none for it.

> I'm not aware about how things are decided around here, but if there are other bash maintainers, it would be interesting to see their input too. If it's possible, I'd like to respectfully ask to reopen this PR.

You got @glennj's input :) My reservations remain. There's better quality material out there. The pure bash is nice but looking at the code, I feel like it's not very high quality.
 
> As I said, the "Pure Bash Bible" seems to be very well aligned with [what we encourage here](https://github.com/exercism/bash/blob/main/exercises/shared/.docs/help.md#external-utilities).

I completely agree it's well aligned. That's not my concern.

kotp:
Also, for some exercises, that resource may even be appropriate to link to for in the mentoring notes.  It is that it is not an appropriate document (even some not so good documentation does have some golden nuggets in there, and that has value) to reference, I just do not know that it is a good document to reference in full, and here.

I refer to students poor material with good nuggets fairly often, knowing that they will then learn the difference, but handing a link off like this as a "useful resource" with the concerns expressed without closer guidance is asking for general problems later on.

PR #565, opened 2022-01-08
fix: absolute path to instructions to install bats

YoginAlex:
Change path to absolute path for instructions to install bats in HELP.md
---

Reviewer Resources:

### DIFF
--- a/exercises/shared/.docs/tests.md
+++ b/exercises/shared/.docs/tests.md
@@ -7,7 +7,7 @@ bats hello_world.bats
 ```
 
 `bats` will need to be installed.
-See the [Testing on the Bash track](/docs/tracks/bash/tests) page for
+See the [Testing on the Bash track](https://exercism.org/docs/tracks/bash/tests) page for
 instructions to install `bats` for your system.
 
 ## Help for assert functions


IsaacG:
The [documentation](https://exercism.org/docs/building/markdown/internal-linking) allows for either style.

What's the impetus/rationale to change this link?

Does this PR catch all relative links in this repo?

kotp:
> What's the impetus/rationale to change this link?

My knee jerk reaction is absolute paths will work 
wherever the markdown file is viewed even when
rendered, while relative will not.

IsaacG:
> My knee jerk reaction is absolute paths will work wherever the markdown file is viewed even when rendered, while relative will not.

That's a good argument which I would agree with ... for making this change across the *entire* repo, and to the upstream documentation policy.

If this PR introduces inconsistencies into this repo and doesn't catch *all* links in the repo, I'd prefer keeping things consistent.

kotp:
On 22/01/10 08:15PM, Isaac Good wrote:
> > My knee jerk reaction is absolute paths will work wherever the
> > markdown file is viewed even when rendered, while relative will not.
> 
> That's a good argument which I would agree with ... for making this
> change across the *entire* repo, and to the upstream documentation
> policy.

My response was only to the quoted portion (concerning what I would
suspect the stimulus was to produce the work here), not to the other point,
both specifically asked, and the others implied.

YoginAlex:
Thanks for the comments!

In my case
- I start a new track - Bash - and a new exercise - Hello world!
- I use "WORK LOCALLY (VIA CLI)" for solve exercises
- I read `HELP.md`
- I know nothing about `bats`
- Link to instructions about `bats` was relative
- So, I can't just click on it and keep reading instruction

As I can see it was only one relative link in `.md` in this repo.

kotp:
I wanted to note [the building guides][1] can be a good way to start, if
you are interested in making the changes repository/track wide or beyond
as well.

[building-guides]: https://exercism.org/docs/building

IsaacG (approved):

IsaacG:
Looks reasonable to me. `grep` suggests this is, in fact, the only relative link in this repo. Approved.

Any reservations, @kotp?

kotp:
On 22/01/11 09:46PM, Isaac Good wrote:
> Looks reasonable to me. `grep` suggests this is, in fact, the only
> relative link in this repo. Approved.
> 
> Any reservations, @kotp?

No reservations.  It makes sense to me to have it work via the link,
even if the file is delivered locally.  (I would rebase and merge, but
it appears that the internet is broken, but of course e-mail works.)

PR #575, opened 2022-04-27
Test for negative arguments

Atomic-Germ:
As it is, the tests don't check for either argument being negative. Negative values being allowed results in unexpected behaviors.
This patch tests for each individually, expecting an error code and message.

---

Reviewer Resources:


IsaacG:
These tests are generated based on the [problems specs data](https://github.com/exercism/problem-specifications/blob/main/exercises/beer-song/canonical-data.json). Please modify tests at the upstream source.

Issue #564, opened 2021-12-31
The `error-handling` instructions could use more bash-specific details.


glennj:
We don't actually specify what we expect the student to do here.

Something like:
> Examine the number of arguments passed to the program.
> If there is exactly one, print a greeting message.
> Otherwise, print a usage message and exit non-zero.

PR #576, opened 2022-05-08
Update instructions for error handling in bash

Philosoft:
The only way to complete this exercise right now is to read tests. PR adds bare bones instructions on what is required

Fixes: #564 

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/error-handling/.docs/instructions.append.md
+++ b/exercises/practice/error-handling/.docs/instructions.append.md
@@ -1,4 +1,6 @@
-# Specific instructions
+# Dummy header
+
+## Bash-pecific instructions
 
 The goal of this exercise is to consider the number of arguments passed to your program.
 If there is exactly one argument, print a greeting message.
--- a/exercises/practice/error-handling/.docs/instructions.append.md
+++ b/exercises/practice/error-handling/.docs/instructions.append.md
@@ -1,0 +1,5 @@
+# Specific instructions
+
+The goal of this exercise is to consider the number of arguments passed to your program.
+If there is exactly one argument, print a greeting message.
+Otherwise print an error message and exit with a non-zero status.
--- a/exercises/practice/error-handling/.docs/instructions.md
+++ b/exercises/practice/error-handling/.docs/instructions.md
@@ -8,3 +8,21 @@ resources even if errors occur.
 This exercise requires you to handle various errors. Because error handling
 is rather programming language specific you'll have to refer to the tests
 for your track to see what's exactly required.
+
+## What you need to do
+
+Implement a script that accepts exactly 1 argument and prints `Hello, <argument>`. If number of arguments is not equals 1, print usage info - `Usage: error_handling.sh <person>` and return non-zero exit code
+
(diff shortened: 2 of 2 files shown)


IsaacG:
> The only way to complete this exercise right now is to read tests.

By and large this is true across multiple tracks and by design, leaning pretty hard into the test-driven development philosophy.

This file is synced directly from the [upstream problem spec](https://github.com/exercism/problem-specifications/blob/main/exercises/error-handling/description.md). **If** we opted to flesh out the instructions, it would need to be done via an amend file.

glennj:
Philosophically speaking, as exercism promotes TDD, getting students to read the tests is exactly what we should be doing. However, I agree that this exercise is a bit under-documented.

As [documented](https://github.com/exercism/docs/blob/main/building/tracks/practice-exercises.md), the place to add this track-specific material is in the `exercises/practice/error-handling/.docs/instructions.append.md` file.

As per exercism's [markdown style guide](https://github.com/exercism/docs/blob/main/building/markdown/markdown.md), all markdown files start with a level-1 header. In this case, the process that joins instructions.md and instructions.append.md drop the header from the append file, so just add a placeholder header.

Regarding the text specifically, I'd prefer the instructions to be a bit less explicit

> The goal of this exercise is to consider the _number_ of arguments passed to your program. If there is exactly one argument, print a greeting message. Otherwise print an error message.

Philosoft:
Ouch. I am sorry. Apparently, I missed mentioning of TDD somewhere. Then maybe there is no need for this change at all 🤔

glennj:
I think this is still a worthy PR, just needs some changes.

Philosoft:
@glennj used your text and moved it to `exercises/practice/error-handling/.docs/instructions.append.md` file

exercises/practice/error-handling/.docs/instructions.append.md
```diff
@@ -0,0 +1,5 @@
+# Specific instructions
```
glennj, line 1: The process that generates README (that concatenates instructions.md and instructions.append.md) strips out level-1 headers.

Please start the file with:
```markdown
# Dummy header

## Bash-specific instructions
```
All of the instructions.append.md files in the bash track are similarly incorrect: I'll create an issue to address that.

glennj (changes requested):

exercises/practice/error-handling/.docs/instructions.append.md
Philosoft: I wasn't sure that the real heading was needed, but 'Dummy header' seemed out of place 😅

Philosoft: https://github.com/exercism/bash/pull/576/commits/0d523b4e5871249198cf82e7aa517b5de46c6f21 ✅

glennj (approved):

glennj:
This is what the instructions look like now:
image

glennj:
Oops, a typo: `Bash-pecific`. we'll need another PR to fix that.

PR #589, opened 2022-07-10
reverse-string: add test with newlines

### LINKED ISSUE #561: command rev is not always working in exercism tests
Hi,

I was implementing https://exercism.org/tracks/bash/exercises/luhn, and I tried to use `rev` to reverse the input string, so I can iterate in a positive loop (starting from 0 index).

While locally this worked fine, when I submitted it, one of the tests executed by exercism.io failed:
```
run bash luhn.sh "055£ 444$ 285"
assert_success
assert_output "false"
```

```
(from function `assert_output' in file bats-extra.bash, line 394,
 in test file luhn.bats, line 94)
  `assert_output "false"' failed

-- output differs --
expected (1 lines):
  false
actual (2 lines):
  rev: stdin: Invalid or incomplete multibyte or wide character
  false
--
```

---

I don't know if it's in the spirit of this track to use such commands, but I thought I should report this regardless :)

glennj:
closes #561

### DIFF
--- a/exercises/practice/reverse-string/.meta/example.sh
+++ b/exercises/practice/reverse-string/.meta/example.sh
@@ -1,3 +1,3 @@
 #!/usr/bin/env bash
 
-echo "$1" | rev
+echo "$1" | rev | tac
--- a/exercises/practice/reverse-string/reverse_string.bats
+++ b/exercises/practice/reverse-string/reverse_string.bats
@@ -62,3 +62,12 @@ load bats-extra
   assert_success
   assert_output "b  * a "
 }
+
+# The program should reverse the entire input, not each line individually
+@test "treats a newline like any other character" {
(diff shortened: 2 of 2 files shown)


IsaacG:
Should this be floated up into the problem spec?

glennj:
Yeah I'll do that. I half suspect it will be rejected there.

The next question that comes to mind: what should awk do with it? A topic for the awk repo.

IsaacG:
I'm fine with sticking to the problem specs. If it's not in the problem spec, then we should not test for it and leave the behavior unspecified.

glennj:
Converted to draft PR so no accidental merges.

IsaacG:
Sounds good to me!

glennj:
Closing

PR #594, opened 2022-09-16
Amended the instructions to explain everything expeccted

maartentroost:
so we dont have to read the bats file....

# pull request template

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/error-handling/.docs/instructions.append.md
+++ b/exercises/practice/error-handling/.docs/instructions.append.md
@@ -3,5 +3,6 @@
 ## Bash-specific instructions
 
 The goal of this exercise is to consider the number of arguments passed to your program.
-If there is exactly one argument, print a greeting message.
-Otherwise print an error message and exit with a non-zero status.
+If there is exactly one argument, print a greeting message: `Hello, argument1`.
+
+Otherwise exit with a non-zero status and print a usage message `Usage: error_handling.sh <person>`.


exercises/practice/error-handling/.docs/instructions.append.md
```diff
@@ -3,5 +3,6 @@
 ## Bash-specific instructions
 
 The goal of this exercise is to consider the number of arguments passed to your program.
-If there is exactly one argument, print a greeting message.
-Otherwise print an error message and exit with a non-zero status.
+If there is exactly one argument, print a greeting message: `Hello, argument1`.
```
glennj, line 6: Should we be more obvious that "argument1" is not a static string? `` `Hello, $argument1` ``

glennj (changes requested):

glennj:
@IsaacG makes a good point.

I didn't read the description before -- "so we dont have to read the bats file" is really antithetical to the TDD approach: you **have to** read the tests to solve the exercises. The tests define the requirements.

If we start down this path, we'll have to change the READMEs for _every_ exercise, duplicating the requirements laid out in the tests. Inevitably, the instructions will get out of sync with the tests, creating chaos.

If the tests are unclear or hard to read, that's a different issue, but I vote to close this PR.

glennj:
Would the instructions be more clear if they used the same wording but with bullets?
```
The goal of this exercise is to consider the number of arguments passed to your program:
- if there is exactly one argument, print a greeting message,
- otherwise print an error message and exit with a non-zero status.
```

glennj:
Would the instructions be easier to read if they used bullets?
```markdown
The goal of this exercise is to consider the number of arguments passed to your program:

* if there is exactly one argument, print a greeting message.
* otherwise print an error message and exit with a non-zero status.
```

glennj:
@maartentroost do you have any more thoughts about this PR?

maartentroost:
The approach on how to solve/work the exercises is indeed the first issue to resolve. For me personally the approach was not clearly defined nor was it apparent in how the exercises are designed. So how should a user approach this? Do we have that approach documented and explained to the user? As mentioned, it wasn’t clear to me. Once we know that, we can adapt everything (documentation, workflow, …) to that model.
I would like to stress that both models use the test driven development approach, they are just slight variations that suit to different contexts.
**Model A)	Tests define and document:**
1. User selects an exercise (no instructions needed, we do need a general description)
2. The tests are ran
3. The error message defines and illustrates the spec to implement
4. User implements
5. Goto 2 until tests succeed

In the “Tests define and document” approach all specifications and instructions are in the tests. No prose instructions are necessary. Step b could run automatically. The initial code must fail the first test. The first test must illustrate clearly the basic premise of the exercise.
**Model B)	Documentation defines and tests illustrate:**
1. User selects an exercise (The instructions should be visible)
2. User implements
3. The tests are ran
4. The error message illustrates the spec to refine
5. Goto 2 until tests succeed

The instructions are in plain English and define all requirements and should be kept in sync with the tests. Tests illustrate the specification that fails but should not go out of their way to explain everything. 
One example of this approach is [https://exercism.org/tracks/bash/exercises/two-fer](https://exercism.org/tracks/bash/exercises/two-fer)

What is our model? A or B? Or C? This is not clear to everyone here involved (at least not to me). Nor is it clearly explained to users (a tutorial this-is-how-you-should-work). This should be made clear and enforced throughout the site and checked for in every exercise in the repositories by the contributors.
When I made some exercises I worked with model B without any problem for several exercises. As an educator I use TDD to steer students, just like we do here. But I write clear instructions to read before implementing. Meaning I use model B. This is, I think, lowering the difficulty of the first step. Reading and comprehension is a well-trained skill to them, reading error messages and extracting goals from these tests are not trained skills (yet).
I wont choose any model now but imho both are viable.

PR #627, opened 2023-04-14
Learning page: bash-hackers.org is defunct

glennj:
sadly.


glennj:
Hang on, I'll use archive.org instead

PR #630, opened 2023-04-28
Update all_your_base.bats

ljsr:
Zero is zero in whatever base, then the correct output for whatever list of zeros must be zero.  For example in the base 10: 0 is zero, to passing this to other base greater of 1 you need divide 0 between the number of the other base: 0 divided on other number have quotient 0 and residue 0 hence 0 from base 10 is 0 in whatever base greater of 1.

# pull request template

---

Reviewer Resources:

[Track Policies](https://github.com/exercism/bash/blob/master/POLICIES.md)

PR #631, opened 2023-04-29
[All Your Base]: Update example.sh

ljsr:
This change permit that the program pass correctly the test related with the digits of zero value:  for example the arguments 10 "0" 2 must return 0 or the arguments 10 "0 0 0 0" 4 must return 0 too.  Additionally this change is accord with the null digit: 10 "" 2, the return value must be 0.

# pull request template

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/all-your-base/.meta/example.sh
+++ b/exercises/practice/all-your-base/.meta/example.sh
@@ -21,6 +21,11 @@ main() {
         decimal=$(( from_base * decimal + digit ))
     done
 
+    if (( $decimal == 0 )); then
+        echo 0
+        return 0
+    fi
+
     digits=()
     while (( decimal > 0 )); do
         digit=$(( decimal % to_base ))


exercises/practice/all-your-base/.meta/example.sh
```diff
@@ -20,7 +20,12 @@ main() {
         validate_digit "$digit" "$from_base"
         decimal=$(( from_base * decimal + digit ))
     done
-
+
```
kotp, line 23:
```suggestion

```

exercises/practice/all-your-base/.meta/example.sh
```diff
@@ -20,7 +20,12 @@ main() {
         validate_digit "$digit" "$from_base"
         decimal=$(( from_base * decimal + digit ))
     done
-
+    
+    if (( $decimal == 0 )); then
+        echo 0
+        return 0
+    fi
+
```
kotp, line 28:
```suggestion

```

kotp (changes requested):
Trailing white space introduced on these two lines should be removed.

glennj:
Please add the updated test case to this PR, and then we can close the other one.

kotp (approved):

glennj:
Work has proceeded in #630

PR #638, opened 2023-09-07
Fix #637: pass actual newlines to the exercise script, not literal '\n' strings

### LINKED ISSUE #637: 'Word Count' test 'handles expanded lists' is mis-executed
Despite 104 people having solved this by bludgeoning the input (https://exercism.org/tracks/bash/exercises/word-count/solutions) -- the test ought to be correct!

Like test #13 'alternating word separators are not detected as a word', test #5 'handles expanded lists' should pass actual newlines to the exercise script; not literal '\n' strings.
```
$ git diff
diff --git a/exercises/practice/word-count/word_count.bats b/exercises/practice/word-count/word_count.bats
index f6c0d4d..e2ba3cb 100644
--- a/exercises/practice/word-count/word_count.bats
+++ b/exercises/practice/word-count/word_count.bats
@@ -45,7 +45,7 @@ load bats-extra
 
 @test "handles expanded lists" {
   [[ $BATS_RUN_SKIPPED == "true" ]] || skip
-  run bash word_count.sh "one,\ntwo,\nthree"
+  run bash word_count.sh $'one,\ntwo,\nthree'
   assert_success
   assert_line "one: 1"
   assert_line "two: 1"
```

filbo:
Not much to say, just run the actual test which was intended...

# pull request template

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/word-count/word_count.bats
+++ b/exercises/practice/word-count/word_count.bats
@@ -45,7 +45,7 @@ load bats-extra
 
 @test "handles expanded lists" {
   [[ $BATS_RUN_SKIPPED == "true" ]] || skip
-  run bash word_count.sh "one,\ntwo,\nthree"
+  run bash word_count.sh $'one,\ntwo,\nthree'
   assert_success
   assert_line "one: 1"
   assert_line "two: 1"


glennj (approved):
Good catch!

filbo:
Thanks, @glennj ...

I noticed that #637, the bug report, showed up on my exercism.com/journey/reputation as a 'Building Exercism' kudo; but #638, the PR, did not.  I'm not complaining, just curious what rules are being followed by the system, and the thoughts behind them.  (There is probably a web link which would immediately answer this...)

IsaacG:
@filbo See https://exercism.org/docs/using/product/reputation and https://exercism.org/docs/building

I believe "building" means "contributing to the exercises in some way" even if that "way" is a tweak to a test input.

glennj:
In your reputation, you _should_ see `You created PR#638 on bash: Fix #637: pass actual newlines to the exercise script, not literal '\n' strings` with 5 reputation points.

filbo:
Yeah, my mistake, it was #638-the-PR which showed up and #637-the-bug which didn't generate any credit.  Which was mildly surprising, but I see now in the guide that only major issue reports get credit.  That's fine, I just needed to find the info to de-puzzle myself :)

PR #650, opened 2024-01-02
Syncx

glennj:
None of the updated tests required any changes to the bats files.

### DIFF
--- a/exercises/practice/queen-attack/.docs/instructions.md
+++ b/exercises/practice/queen-attack/.docs/instructions.md
@@ -8,18 +8,14 @@ A chessboard can be represented by an 8 by 8 array.
 
 So if you are told the white queen is at `c5` (zero-indexed at column 2, row 3) and the black queen at `f2` (zero-indexed at column 5, row 6), then you know that the set-up is like so:
 
-```text
-  a b c d e f g h
-8 _ _ _ _ _ _ _ _ 8
-7 _ _ _ _ _ _ _ _ 7
-6 _ _ _ _ _ _ _ _ 6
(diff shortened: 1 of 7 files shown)


exercises/practice/queen-attack/.docs/instructions.md
```diff
@@ -8,18 +8,14 @@ A chessboard can be represented by an 8 by 8 array.
...
-2 _ _ _ _ _ B _ _ 2
-1 _ _ _ _ _ _ _ _ 1
-  a b c d e f g h
-```
+![A chess board with two queens. Arrows emanating from the queen at c5 indicate possible directions of capture along file, rank and diagonal.](https://assets.exercism.org/images/exercises/queen-attack/queen-capture.svg)
 
 You are also able to answer whether the queens can attack each other.
 In this case, that answer would be yes, they can, because both pieces share a diagonal.
+
+## Credit
```
kotp, line 16: Should this be plural?

kotp (approved):
Approving with only one non-blocking question.

IsaacG (approved):

exercises/practice/queen-attack/.docs/instructions.md
glennj: Not necessarily: "credit where credit is due"

kotp: That is an example wording where it iterates individually, though, as I understand the language.  The credits are shown at the end of the movie as a collection, such as we do here.

Either way, I think it is a non-blocker.

PR #657, opened 2024-01-21
Approaches for reverse-string

glennj:
Plus a performance article.

### DIFF
--- a/exercises/practice/reverse-string/.approaches/external-tools/content.md
+++ b/exercises/practice/reverse-string/.approaches/external-tools/content.md
@@ -1,0 +1,27 @@
+# External Tools
+
+The GNU Linux core utilities contain lots of goodies.
+To reverse strings, use `rev`.
+
+```bash
+$ echo "Hello, World!" | rev
+!dlroW ,olleH
+```
+
+`rev` also works with files to reverse each line.
+
+```bash
+$ printf '%s\n' one two three > myfile
+$ rev myfile
+eno
+owt
+eerht
+```
+
+There are other ways to do this, but none are a simple as `rev`.
+
+```bash
+echo "$string" | grep -o . | tac | paste -s -d ''
+echo "$string" | perl -lne 'print scalar reverse'
+# etc
+```
--- a/exercises/practice/reverse-string/.approaches/external-tools/content.md
+++ b/exercises/practice/reverse-string/.approaches/external-tools/content.md
@@ -21,7 +21,7 @@ eerht
 There are other ways to do this, but none are a simple as `rev`.
 
 ```bash
-echo "$string" | grep -o . | tac | paste -s -d ''
-echo "$string" | perl -lne 'print scalar reverse'
+grep -o . <<< "$string" | tac | paste -s -d ''
+perl -lne 'print scalar reverse' <<< "$string"
 # etc
 ```
--- a/exercises/practice/reverse-string/.approaches/introduction.md
+++ b/exercises/practice/reverse-string/.approaches/introduction.md
@@ -37,7 +37,7 @@ For more details, go to the [Looping approach][app-loop].
 ## Which approach to use?
 
 Calling out to `rev` makes this exercise extremely trivial.
-In a work environment, it's exactly the approach to take.
+In a production environment, it's exactly the right approach to take.
 
 If you're interested in learning about bash loops and parameter expansion, experimenting with the loop approach is more interesting.
 
--- a/exercises/practice/reverse-string/.approaches/introduction.md
+++ b/exercises/practice/reverse-string/.approaches/introduction.md
@@ -1,0 +1,30 @@
+
+  167  str='Hello, World!'
+  168  rev <<< $str
+  169  grep -o . <<< $str
+  170  grep -o . <<< $str | tac
+  171  grep -o . <<< $str | tac | paste -s -d ''
+  172  awk -F '' '{for (i = 1, j = NF; i < j; i++, j--) {tmp = $i; $i = $j; $j = tmp}; print $i}' <<< $str
+  173  awk -F '' '{for (i = 1, j = NF; i < j; i++, j--) {tmp = $i; $i = $j; $j = tmp}; print}' <<< $str
+  174  reverse() { local reversed=""; local i; for ((i = 0; i < ${#1}; i++)); do reversed=${1:i:1}$reversed; done; echo "$reversed"; }
+  175  reverse "$str"
+  176  set +x
+  177  declare -f reverse
+  178  set +x
+  179  reverse "$str"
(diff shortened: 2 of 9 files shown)


exercises/practice/reverse-string/.approaches/introduction.md
```diff
@@ -0,0 +1,50 @@
...
+    reversed+="${string:i:1}"
+done
+```
+
+For more details, go to the [Looping approach][app-loop].
+
+## Which approach to use?
+
+Calling out to `rev` makes this exercise extremely trivial.
+In a work environment, it's exactly the approach to take.
```
IsaacG, line 40:
```suggestion
In a production environment, it's exactly the right approach to take.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
+# Performance
+
+We usually don't care that much about performance with shell scripts.
+But when we can avoid doing something we know is slow, especially if this slow thing is done repeatedly inside a loop, avoid it.
```
IsaacG, line 4:
```suggestion
When we know something is slow, we can avoid it.
This matters much more when the operation is done inside loop.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+
+$ len=${#string}
+$ time for ((i = 0; i < len; i++)); do :; done
+
+real    0m0.159s
+user    0m0.159s
+sys     0m0.000s
+```
+
+If we can loop backwards, don't even need the variable.
```
IsaacG, line 30:
```suggestion
If we can loop backwards, we don't even need to save the length to a variable.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+We even get a tiny improvement since bash does not need to access the variable contents for each iteration.
```
IsaacG, line 31:
```suggestion
We get a tiny improvement since bash does not need to access the variable contents for each iteration.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+$ time for ((i = ${#string} - 1; i >= 0; i--)); do true; done
+
+real    0m0.151s
+user    0m0.151s
+sys     0m0.000s
+```
+
+## Substrings can be slow
+
+Whether we go backwards or forwards, we are still having to extract the character at the given index.
```
IsaacG, line 43:
```suggestion
Whether we go backwards or forwards, we still have to extract the character at the given index.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+Re-using our 32,767 character string:
+
+```bash
+$ time for ((i = 0; i < len; i++)); do
+    char="${string:i:1}"
+  done
```
IsaacG, line 49:
```suggestion
done
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+sys     0m0.060s
+```
+
+There's a 27x improvement.
+
+What are we doing there?
+
+- `< <(printf "%s" "%string)` is redirecting (`<`) a [process substitution][process-subst] (`<(...)`).
+    We could use a [here-string][here-string] (`<<< "$string"`), but that appends a newline.
+    Using `printf` like that redirects the string exactly as it is.
```
IsaacG, line 80:
```suggestion
    Using `printf` outputs the string without adding a trailing newline.
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+- `IFS= read -d "" -r -n 1 char` OK there's a lot going on there with this [`read` command][read]:
```
IsaacG, line 81:
```suggestion
- `IFS= read -d "" -r -n 1 char` there's a lot going on there with this [`read` command][read]:
```

exercises/practice/reverse-string/.articles/performance/content.md
```diff
@@ -0,0 +1,101 @@
...
+    - `IFS=`: Normally, `read` will trim leading and trailing whitespace.
+        More exactly, characters in `$IFS` are removed from the start and end of the text that `read` captures.
+        By default, IFS contains space, tab and newline.
+        Our goal is to read each character of the string, including whitespace.
+        That empty assignment assigns the empty string to IFS _only for the duration of the `read` command_ so that whitespace characters are not treated specially.
```
IsaacG, line 86:
```suggestion
        The empty assignment assigns the empty string to IFS _only for the duration of the `read` command_ so that whitespace characters are not treated specially.
```

exercises/practice/reverse-string/.articles/performance/snippet.md
```diff
@@ -0,0 +1,5 @@
+```bash
+while IFS= read -d "" -r -n 1 char; do
+    echo "$char"
```
IsaacG, line 3:
```suggestion
    echo -n "$char"
```

exercises/practice/reverse-string/.approaches/external-tools/content.md
```diff
@@ -0,0 +1,27 @@
+# External Tools
+
+The GNU Linux core utilities contain lots of goodies.
+To reverse strings, use `rev`.
+
+```bash
+$ echo "Hello, World!" | rev
```
IsaacG, line 7: Is there any reason you're not using a herestring here?

exercises/practice/reverse-string/.articles/performance/content.md
glennj: I'll use
```markdown
- `IFS= read -d "" -r -n 1 char`: There's a lot going on there with this [`read` command][read].
```

exercises/practice/reverse-string/.articles/performance/content.md
glennj: I'll leave it as is.

exercises/practice/reverse-string/.approaches/external-tools/content.md
glennj: Because using `echo` is probably most familiar to people.

I'll change the weirder examples at the bottom to use herestrings. The `grep | tac | paste` one will be more interesting with a herestring due to its location.

IsaacG (approved):

PR #664, opened 2024-02-17
Add exercise Circular Buffer

glennj:
I had a thought about how to implement this. It's a library like the list-ops exercise.

I'm curious what you all think.

Because it's a library, the tests don't use the bats `run` command, so `assert_success` and `assert_failure` don't apply. You'll see test assertions look like this
```bash
    buffer::write buff 'A'      && result=ok || result=fail
    assert_equal "$result" ok
```

I didn't write any additional instructions. It's a fairly difficult exercise, so users are on their own.

### DIFF
--- a/exercises/practice/circular-buffer/.meta/example.sh
+++ b/exercises/practice/circular-buffer/.meta/example.sh
@@ -1,0 +1,94 @@
+#!/usr/bin/env bash
+
+# Namerefs require bash version 4.3 or greater
+bash_version=$((10 * BASH_VERSINFO[0] + BASH_VERSINFO[1]))
+if (( bash_version < 43 )); then
+    echo "This library requires at least bash version 4.3" >&2
+    return 4
+fi
+
+# I'm storing all the data for all circular buffers in a single global
+# associative array.
+# Empty cells are denoted by empty strings.
+# Therefore, an empty string cannot be stored in this circular buffer
+
+declare -gA _buffer_data=()
+
+buffer::new() {
+    local name=$1
+    local size=$2
+    _buffer_data[size,$name]=$size
+    buffer::clear "$name"
+}
+
+buffer::destroy() {
+    local name=$1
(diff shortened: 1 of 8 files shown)


kotp (approved):

kotp:
Get it out there and see what falls from the sky‽

glennj:
I had a thought just now. What if, instead of 
```sh
buffer::overwrite bufname value
```
we had
```sh
buffer::write -f bufname value
```
That would get the student to to option handling inside the function.

That's a more shell-like syntax, but adds complexity. 

Worth it?

kotp:
Is this the exercise we want to introduce additional complexity in?  It is already admittedly somewhat difficult.

exercises/practice/circular-buffer/.meta/example.sh
```diff
@@ -0,0 +1,102 @@
...
+
+    buffer::incr_pointer "$name" read
+}
+
+buffer::write() {
+    local overwrite=false
+    # To use `getopts` in a function, these vars must be local.
+    local OPTIND OPTARG
+    while getopts :f opt; do
+        case $opt in
```
IsaacG, line 74:
```suggestion
        case "$opt" in
```

IsaacG (approved):
The PR contents look good to me. The idea behind it...

I'm a bit iffy here based on the complexity. Just because you can get bash to do this doesn't mean it's reasonable to do so. I'm leaning more towards passing on exercises that don't fit the language vs jumping through hoops to get things working.

glennj:
> I'm a bit iffy here based on the complexity. Just because you can get bash to do this doesn't mean it's reasonable to do so. I'm leaning more towards passing on exercises that don't fit the language vs jumping through hoops to get things working.

That is completely reasonable. I'm content to close this.

I'll create an issue so we can discuss foregoing exercises.

glennj:
Although this is do-able in bash, it doesn't mean that it's smart to do it.

PR #668, opened 2024-02-28
Add a test which fails if lexicographical sorting was used

knight-erraunt:
Add test verifying that the sorting happened in a numerical fashion and not lexicographical.

---

Reviewer Resources:

### DIFF
--- a/exercises/practice/triangle/triangle.bats
+++ b/exercises/practice/triangle/triangle.bats
@@ -153,7 +153,7 @@ load bats-extra
   assert_output "false"
 }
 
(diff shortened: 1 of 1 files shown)


IsaacG:
Please see https://exercism.org/docs/community/being-a-good-community-member/suggesting-exercise-improvements which has a couple of relevant points which apply here.

There's also metafiles which need updating when tests change. You can read about those here: https://exercism.org/docs/building/tracks/practice-exercises#h-file-meta-tests-toml

PR #672, opened 2024-03-28
sync change exercise

glennj:


glennj:
> adds an edge case that is not worth rerunning all tests for

I'd say this applies

IsaacG (approved):

IsaacG:
This checks for faulty algorithms. I would think rerunning tests would be wanted here.

PR #681, opened 2024-04-27
Add minesweeper exercise

glennj:


exercises/practice/minesweeper/.meta/example.sh
```diff
@@ -0,0 +1,51 @@
...
+
+# some global vars
+declare -A board
+height=$#
+width=${#1}
+
+parse_input() {
+    local rownum=0 row col index char
+    for row in "$@"; do
+        for ((col=0; col < "${#row}"; col++)); do
```
IsaacG, line 11: Is there a reason you space pad the `<` operator but not the `=` operator?

IsaacG (approved):

exercises/practice/minesweeper/.meta/example.sh
kotp: I, of course, cannot answer this question for the author, but it does seem to be consistent with the other assignments such as line 5 and 6.  I am all for space around operators where allowed by the language though, as it makes the "words", well, words.

exercises/practice/minesweeper/minesweeper.bats
```diff
@@ -0,0 +1,189 @@
...
+}
+
+@test "mine surrounded by spaces" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    input=(
+        "   "
+        " * "
+        "   "
+    )
+    run bash minesweeper.sh "${input[@]}"
```
glennj, line 59: My main concern was how to represent an array of strings, particularly whitespace-only strings, as input to the script. This method OK with everyone @here?

IsaacG: It gets the job done so no objections here. You can also use STDIN and `printf "%s\n" "${data[@]}" | ...`.

IsaacG (approved):

exercises/practice/minesweeper/minesweeper.bats
glennj: I did try that, thought it was too messy.

PR #700, opened 2024-08-16
Concept: variables

glennj:
Next concept: variables

This one is much more terse. I may have swung the pendulum too far.


concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
+# Introduction
+
+## Variables
+
+Variables are a place to hold data.
+Bash has two kinds of data: strings and arrays.
+We'll talk about arrays later.
+
+You assign a variable like this
```
IsaacG, line 9:
```suggestion
You assign a variable (often shortened to var) like this:
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+```
+
+bash is going to call the **command** `x` with the **two arguments** `=` and `10`!
+~~~~
+
+## Parameters
+
+Variables can also be called "parameters".
+The difference is that a variable is assigned by you, the programmer.
+A parameter can be a read-only variable that is set by the shell.
```
IsaacG, line 30:
```suggestion
Many variables are assigned by you, the programmer.
Some variables are set automatically by the shell.
These are called parameters.
```

As far as I know, bash doesn't many any of those read only.

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+
+Some important parameters used by or set by bash include:
+
+* `PATH` -- a colon-separated list of directories used to find external commands.
+* `IFS` -- the "internal field separator". We'll see how it's used below.
```
IsaacG, line 35:
```suggestion
* `IFS` -- the "internal field separator". 
   We'll see how it's used below.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+* `PWD` -- your current working directory.
```
IsaacG, line 36:
```suggestion
* `PWD` -- your current (or present) working directory.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+* `BASH_VERSION` -- the version string of the running instance of bash.
+
+## Parameter Names
+
+A variable name consists of uppercase letters, lowercase letters, numbers and underscore.
+It cannot start with a number.
+
+~~~~exercism/caution
+It is best practice to **avoid using ALLCAPS variable names**.
+The shell uses this convention for its parameters, and you don't want to accidentally override shell parameters that have special meaning.
```
IsaacG, line 47:
```suggestion
The shell uses this convention for its parameters, and you don't want to accidentally overwrite shell parameters that have special meaning.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+For example: [1][1], [2][2], [3][3]
+
+[1]: https://stackoverflow.com/q/27555060/7552
+[2]: https://stackoverflow.com/q/28310594/7552
+[3]: https://unix.stackexchange.com/q/114596/4667
+~~~~
+
+## Parameter Expansion
+
+You get the value of a parameter with the `$varname` syntax.
```
IsaacG, line 57: Did you just define "parameters" are variables set by the shell? That would imply PE only applies to shell vars.

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+To prevent the variable name from being confused with surrounding text, you can enclose the variable name in braces.
+For example to print the string `10x10`
+
+```bash
+echo "${x}x$x"
+# or
+echo "${x}x${x}"
+```
+
+Some [style guides][google-style-guide] recommend using braces in most cases for readability.
```
IsaacG, line 73:
```suggestion
Some [style guides][google-style-guide] recommend using braces in most cases for readability and/or consistency.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+
+## Positional Parameters
+
+As we discussed in the [Commands and Arguments][cmds-args] lesson, commands can take arguments.
+Bash scripts are no exception.
+The arguments to the current instance of bash are called "positional parameters".
```
IsaacG, line 79: If you're inside a function, the positional params change and are not the args to the current shell instance. I don't know if it's worth making that distinction or not here.

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+
+There are some "special" parameters that help you work with the positional parameters.
+
+* `$#` expands to the _number_ of positional parameters.
+* `"$@"` expands to the list of all of the positional parameters, each as a separate word.
+* `"$*"` expands to a single string of all the positional parameters joined together by a character.
+  * The join character is _the first character_ of the `$IFS` variable, which is a space by default.
+
+The `"$@"` is the safest way to pass positional parameters to another command while keeping them safely quoted.
+It is also what you will use to loop over them:
```
IsaacG, line 97:
```suggestion
It is also what you will use to loop over the positional parameters.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+
+```bash
+for arg in "$@"; do
+    do_something_with "$arg"
+done
+```
+
+## Command Substitution
+
+A very frequent operation you will do in bash scripts is to capture the output of a command and store it in a variable.
```
IsaacG, line 107: This doc mentions "you" a whole lot and things "you will do". I think the prior doc was a little bit less ...directed?

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+To do this, you will use the command substitution syntax, like this:
```
IsaacG, line 108:
```suggestion
To do this, you will use _command substitution_.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+uppercase=$(echo "$text" | tr '[:lower:]' '[:upper:]')
+echo "$uppercase"
+# => HELLO WORLD!
+```
+
+## More on Parameter Expansion
+
+Bash has many builtin facilities to manipulate variables and strings so that you don't need to call out to external commands.
+
+* Uppercasing text: in the previous section I showed how to use `tr` to upper case some text. To do it in bash:
```
IsaacG, line 127:
```suggestion
* Uppercasing text. The previous section showed how to use `tr` to upper case some text. Bash allows you to do that without using `tr`.
```

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+
+* Search and replace.
+
+  ```bash
+  text="Hello world!"
+  echo "${text//[aeiou]/X}"
+  # => HXllX wXrld!"
+  ```
+
+There are many other interesting parameter expansions.
```
IsaacG, line 150: I wonder if it's worth at least mentioning a bunch of the others.

concepts/variables/introduction.md
```diff
@@ -0,0 +1,156 @@
...
+* `${10}` ... -- double digits require braces.
+* `$0` -- the name of the current script.
+
+### Special Positional Parameters
+
+There are some "special" parameters that help you work with the positional parameters.
+
+* `$#` expands to the _number_ of positional parameters.
+* `"$@"` expands to the list of all of the positional parameters, each as a separate word.
+* `"$*"` expands to a single string of all the positional parameters joined together by a character.
```
IsaacG, line 93: Should `$*` have a caveat that it is usually not the right one to use?

concepts/variables/introduction.md
glennj: Is it often shortened to var? That's news to me.

concepts/variables/introduction.md
glennj: How about just this?
```
Variables can also be called "parameters".
The two terms will be used interchangably in this document.
```

concepts/variables/introduction.md
glennj: Well, the first doc was not written by me. Maybe we can "put a pin in it" and revisit it after a few docs are written.

concepts/variables/introduction.md
IsaacG: Yes. I mean, your next line is `varname` and not `variablename` ;) Or you can avoid shortening it in the example.

concepts/variables/introduction.md
IsaacG: I could modify the factoid messages and/or Wooledge wiki, if that helps :D

IsaacG (approved):

concepts/variables/introduction.md
glennj: resolved this by using `name="value"` for the example.

concepts/variables/introduction.md
glennj: Upon reflection, I'll circle back to the first concept to give it a rewrite in my voice.

concepts/commands-and-arguments/links.json
```diff
@@ -1,10 +1,10 @@
 [
   {
-    "description": "Commands and Arguments in the Bash Guide",
+    "description": "\"Commands and Arguments\" in the BashGuide",
```
kotp, line 3: In prose should "Bash Guide" be two words?   That is why in code we would CamelCase them, right?

concepts/variables/introduction.md
kotp: For teaching material, regardless, I do prefer using full names, even if it is often shortened in practice.  Those that are not "acronyms and abbreviations as a first language" can then easily look up those words in an English dictionary, making interpretation and deciphering easier (especially on top of mult-language audience/readers being highly likely on this platform).

kotp (approved):
Knowing that some rewrite to different voice is pending...

concepts/commands-and-arguments/links.json
glennj: It's literally a WikiWord, but that's not no common anymore.

glennj:
https://exercism.org/tracks/bash/concepts/variables

PR #703, opened 2024-08-27
 Concept quoting

glennj:
I'm using [admonitions](https://exercism.org/docs/building/markdown/markdown#h-special-blocks-sometimes-called-admonitions) pretty liberally here, so we won't really see exactly how they get rendered until the PR is merged.

### DIFF
--- a/concepts/quoting/introduction.md
+++ b/concepts/quoting/introduction.md
@@ -117,7 +117,8 @@ However, leaving variables unquoted is not recommended unless you also disable F
 ### Filename Expansion
 
 "Glob" patterns are used to express a concise pattern to match a set of files.
-For example, you might match all the TXT files in a directory with `*.txt`. [Glob patterns][glob] have existed since the invention of Unix.
+For example, you might match all the Markdown files in a directory with `*.md`.
+[Glob patterns][glob] have existed since the invention of Unix.
 
 The glob wildcard characters are:
 
@@ -127,16 +128,18 @@ The glob wildcard characters are:
 
    ~~~~exercism/advanced
    <details><summary>
-   There are lots of details about bracket expressions.
+   There are lots of details about character sets.
    Click here to show them.
    </summary>
 
     * `[abc]` matches exactly one character that is either an `a` or a `b` or a `c`.
         * to include a literal close bracket in the expression, make it the first character after the opening bracket (after an optional `^`)
-            * `[][]` is a bracket expression that will match either a close or an open bracket.
+            * `[][]` is a character set that will match either a close or an open bracket.
+            * or escape it with a backslash: `[[\]]`
     * `[x-y]` matches exactly one character that is in the _range_ from `x` to `y`
         * to include a literal hyphen character in a character set, it must be either the first character after the opening bracket (after an optional `^`) or the last character before the closing bracket.
             * `[-xy]` or `[xy-]` each match exactly one character that is an `-` or a `x` or a `y`
+            * or escape it with a backslash: `[x\-y]`
         * multiple ranges can be specified, for example `[0-9A-Fa-f]` matches a hexadecimal digit
     * `[^abc]` matches exactly one character that is **NOT** an `a` or a `b` or a `c`,
     * `[^x-y]` matches exactly one character that is **NOT** in the range from `x` to `y`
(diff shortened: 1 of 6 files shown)


concepts/quoting/introduction.md
```diff
@@ -0,0 +1,282 @@
...
+```none
+1 The quick brown fox jumps.
+```
+
+However, leaving variables unquoted is not recommended unless you also disable Filename Expansion.
+
+### Filename Expansion
+
+"Glob" patterns are used to express a concise pattern to match a set of files.
+For example, you might match all the TXT files in a directory with `*.txt`. [Glob patterns][glob] have existed since the invention of Unix.
```
IsaacG, line 120:
```suggestion
For example, you might match all the text files in a directory with `*.txt`.
[Glob patterns][glob] have existed since the invention of Unix.
```

concepts/quoting/introduction.md
```diff
@@ -0,0 +1,282 @@
...
+
+The glob wildcard characters are:
+
+* `*` matches zero or more of any character,
+* `?` matches exactly one of any character,
+* `[...]` matches exactly one character from the set of characters listed inside the brackets.
+
+   ~~~~exercism/advanced
+   <details><summary>
+   There are lots of details about bracket expressions.
```
IsaacG, line 130: I haven't heard these calls bracket expressions before. I usually call them a character set, though that might be taken from the regex world.

IsaacG (approved):

concepts/quoting/introduction.md
glennj: It's a term I got [from Tcl](https://www.tcl.tk/man/tcl8.6/TclCmd/re_syntax.htm#M30). There doesn't seem to be a consensus, even among regex terminologies:

* Ruby: [character classes](https://ruby-doc.org/3.3.4/Regexp.html#class-Regexp-label-Character+Classes) -- this I actively want to avoid, since bash calls `[:alpha:]` a "character class" within a character set.
* Python regex: [a set of characters](https://docs.python.org/3/library/re.html#index-10)
* Python glob: I can't see it described in the glob or fnmatch modules.
* Perl: File::Glob calls it a [character class](https://perldoc.pl/File::Glob#META-CHARACTERS)

I'll change it to "character set". I did a bash mentoring session for acronym where the mentee used `IFS=" ,-,_,*,?"` so describing it as a set of characters might help with that.

glennj:
@IsaacG @kotp bigger picture: Does it make sense to talk about Word Splitting and Filename Expansion in the context of Quoting?

concepts/quoting/introduction.md
kotp: Referenced from https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap09.html

```
The following character class expressions shall be supported in all locales:

[:alnum:]   [:cntrl:]   [:lower:]   [:space:]
[:alpha:]   [:digit:]   [:print:]   [:upper:]
[:blank:]   [:graph:]   [:punct:]   [:xdigit:]
```

So perhaps "character class expression" to describe `[:alpha:]`.

kotp:
> @IsaacG @kotp bigger picture: Does it make sense to talk about Word Splitting and Filename Expansion in the context of Quoting?

I think at least filename expansion.  Not sure about word splitting as much as avoiding splitting strings as separate words.

IsaacG:
> @IsaacG @kotp bigger picture: Does it make sense to talk about Word Splitting and Filename Expansion in the context of Quoting?

Seems reasonable to me! Quoting is only needed to avoid word splitting and pattern expansions so they go hand in hand.

> Not sure about word splitting as much as avoiding splitting strings as separate words.

Unless I'm missing something, those two are saying the exact same thing.

glennj:
@kotp have I missed anything that is blocking your approval?

concepts/quoting/introduction.md
```diff
@@ -0,0 +1,285 @@
...
+i=0
+for word in $sentence; do echo "$((++i)) $word"; done
+# .........^.........^
+```
+outputs
+```none
+1 The quick brown fox jumps.
+```
+
+However, leaving variables unquoted is not recommended unless you also disable Filename Expansion.
```
kotp, line 115: Might mention "how to disable" here, or give a reference for further reading.

kotp (approved):
Approving (not sure it will count though) with comment.

PR #707, opened 2024-09-09
Update hamming.bats

pedrobotolli:
A test was added to fail if the script accepts more than two arguments, because the exercise description specifies that exactly two arguments of equal length must be passed.

# pull request template

---

Reviewer Resources:


glennj:
Let's discuss potential changes to the track [on the forum](https://forum.exercism.org/c/programming/bash/70) please, before you raise a PR.

PR #708, opened 2024-09-23
test: update Darts exercise test case

kirillmorozov:
I was able to pass this exercise using incorrect `[[:digit:]]` regular expression that only checked that an argument includes a number so I've decided to slightly update the test case for a non-numeric argument.

---

Reviewer Resources:


glennj:
As someone who recently completed the exercise, was it clear what the test expected?
```
    assert_failure
    assert_output    # there is _some_ output
```

glennj (approved):

kirillmorozov:
Yeah, I quickly figured out that it expects a non-zero exit code and a message without enforcing concrete error message.

PR #712, opened 2024-11-18
Concept: Conditionals

glennj:
I've been kind of agonizing over this one for a few weeks: how much detail to include/omit.

### DIFF
--- a/concepts/conditionals/introduction.md
+++ b/concepts/conditionals/introduction.md
@@ -83,7 +83,7 @@ Some typical conditional expressions include:
 [ "$num1" -eq "$num2" ]        # arithmetic comparisons
 ```
 
-There are many more operations available: they are listed in the [Bash Conditional Expressions][cond-expr] section of the manual.
+There are many more operations available; they are listed in the [Bash Conditional Expressions][cond-expr] section of the manual.
 
 ~~~~exercism/note
 In the examples above, notice that all the variables are quoted.
@@ -114,7 +114,7 @@ if [ -n $str ]; then echo "not empty"; else echo "empty"; fi
 ## "[[" Keyword
 
 The [`[[...]]` conditional construct][cond-construct] is not a _command_, it is a **keyword**.
-This means that, although it is handled like any other command, it can have special parsing rules.
+This means that, although it is handled like any other command, it has special parsing rules.
 What's special about `[[` is that the variables expanded within it are **not** subject to word splitting or filename expansion.
 That means this command acts as you expect, even without quoting.
 
@@ -122,7 +122,7 @@ That means this command acts as you expect, even without quoting.
 if [[ -n $str ]]; then echo "not empty"; else echo "empty"; fi
 ```
 
-`[[` works with all the conditional expressions that `test` and `[` can handle.
+`[[` supports all the conditional expressions that `test` and `[` can handle.
 In addition, `[[` provides
 
 * the `=~` regular-expression matching operator,
--- a/concepts/conditionals/introduction.md
+++ b/concepts/conditionals/introduction.md
@@ -148,23 +148,22 @@ When one matches, the COMMANDS are executed.
 ```bash
 read -p "Guess the secret word: " word
 case "$word" in
-    secret)
-        echo "Yes, you guessed it!"
-        ;;
-    s*)
-        echo "You guessed the first letter."
-        ;;
-    *)
-        echo "Not even close! Try again."
-        ;;
+    secret) echo "Yes, you guessed it!" ;;
+    ??????) echo "That's the right number of letters." ;;
+    s*)     echo "You guessed the first letter." ;;
+    *)      echo "Not even close! Try again." ;;
 esac
 ```
 
 Each COMMANDS clause must end with **two** semicolons, `;;`.
 
 ~~~~exercism/note
-There are alternatives that provide functionality for fall-through command execution, and for pattern matching to continue.
-Check the manual for details.
+1. There are alternatives that provide functionality for fall-through command execution, and for pattern matching to continue.
+   Check [the manual][case] for details.
+2. Recall that we talked about patterns in the [Filename Expansion][glob] section of the Quoting concept.
+
+[case]: https://www.gnu.org/software/bash/manual/bash.html#index-case
+[glob]: https://exercism.org/tracks/bash/concepts/quoting#h-filename-expansion
 ~~~~
 
 [command-list]: https://www.gnu.org/software/bash/manual/bash.html#Lists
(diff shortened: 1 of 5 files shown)


concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
+# Introduction
+
+## Truthiness in Bash
+
+Bash does not have the concept of boolean values.
+There are strings and numbers, and arrays of strings or numbers.
+So how do conditional commands deal with true and false?
+
+## Exit Statuses
```
IsaacG, line 9:
```suggestion
## Exit Status
```

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+Commands produce an _exit status_ when they end.
+An exit status is an integer between 0 and 255 inclusive.
+Bash considers a **zero** exit status to represent _success_.
+Any other exit status represents _failure_.
+
+This applies to bash builtin commands and keywords, as well as functions.
```
IsaacG, line 16:
```suggestion
This applies to all commands, including bash builtin commands, keywords, and functions.
```

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+## The "if" Command
+
+The basic syntax of the `if` command is
+
+```bash
+if CONDITIONAL_COMMANDS; then TRUE_COMMANDS; else FALSE_COMMANDS; fi
```
IsaacG, line 23: You may want to mention the `else` is optional

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+It starts with `if`; the condition commands are separated from the "success" commands by `then`; and it is terminated with `fi`.
+The `else` clause is optional.
+
+CONDITIONAL_COMMANDS can be a single command or it can be a [list of commands][command-list].
```
IsaacG, line 29: Is it worth mentioning that generally the last command in the list sets the exit status of the list?

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+`[[` works with all the conditional expressions that `test` and `[` can handle.
+In addition, `[[` provides
+
+* the `=~` regular-expression matching operator,
+* `==` and `!=` operate as a glob-pattern matching operator,
+* `&&` and `||` as logical operators (special parsing rule),
+* `<` and `>` as "bare" string comparison operators (special parsing rule: because these are redirection symbols, in `[` they must be escaped).
+
+It is the author's opinion that these special features offer so much benefit that `[[` should be used exclusively.
```
IsaacG, line 133: It's the bash community's opinion, too :)

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+```bash
+case WORD in [PATTERN [| PATTERN]...) COMMANDS ;;]... esac
+```
+
+The WORD is matched against each PATTERN.
+When one matches, the COMMANDS are executed.
+
+```bash
+read -p "Guess the secret word: " word
+case $word in
```
IsaacG, line 149:
```suggestion
case "$word" in
```

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+    s*)
+        echo "You guessed the first letter."
+        ;;
+    *)
+        echo "Not even close! Try again."
+        ;;
+esac
+```
+
+Each COMMANDS clause must end with **two** semicolons `;;`.
```
IsaacG, line 162:
```suggestion
Each COMMANDS clause must end with **two** semicolons, `;;`.
```

IsaacG (approved):

concepts/conditionals/introduction.md
glennj: This is an annoying one. Like variable assignments, no word splitting or file expansion is done there. But it looks wrong when it's bare.

concepts/conditionals/introduction.md
glennj: Unless a previous command fails. I think this can be saved until the pipelines/lists concept.

concepts/conditionals/introduction.md
glennj: I do, two sentences later.

concepts/conditionals/introduction.md
IsaacG: The exit status of previous commands only matters when `set -e` is enabled :) But sounds good.

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,174 @@
...
+Within `[` and `]`, you write a conditional expression.
+Some typical conditional expressions include:
+
+```bash
+[ -f "$filename" ]             # file operations
+[ "$string1" = "$string2" ]    # string comparisons
+[ "$num1" -eq "$num2" ]        # arithmetic comparisons
+```
+
+There are many more operations available: they are listed in the [Bash Conditional Expressions][cond-expr] section of the manual.
```
IsaacG, line 86:
```suggestion
There are many more operations available; they are listed in the [Bash Conditional Expressions][cond-expr] section of the manual.
```

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,174 @@
...
+if [ -n $str ]; then echo "not empty"; else echo "empty"; fi
+```
+
+[test]: https://www.gnu.org/software/bash/manual/bash.html#index-test
+~~~~
+
+## "[[" Keyword
+
+The [`[[...]]` conditional construct][cond-construct] is not a _command_, it is a **keyword**.
+This means that, although it is handled like any other command, it can have special parsing rules.
```
IsaacG, line 117:
```suggestion
This means that, although it is handled like any other command, it has special parsing rules.
```

concepts/conditionals/introduction.md
```diff
@@ -0,0 +1,174 @@
...
+What's special about `[[` is that the variables expanded within it are **not** subject to word splitting or filename expansion.
+That means this command acts as you expect, even without quoting.
+
+```bash
+if [[ -n $str ]]; then echo "not empty"; else echo "empty"; fi
+```
+
+`[[` works with all the conditional expressions that `test` and `[` can handle.
```
IsaacG, line 125:
```suggestion
`[[` supports all the conditional expressions that `test` and `[` can handle.
```

IsaacG (approved):
Looks good!

glennj:
Visible at https://exercism.org/tracks/bash/concepts/conditionals

kotp:
Looks really good!

PR #715, opened 2024-12-20
Concept: functions

glennj:
Functions concept

### DIFF
--- a/concepts/functions/.meta/config.json
+++ b/concepts/functions/.meta/config.json
@@ -1,0 +1,10 @@
+{
+  "authors": [
+    "glennj"
+  ],
+  "contributors": [
+    "kotp",
+    "IsaacG"
+  ],
+  "blurb": "Functions in Bash programs."
+}
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -1,0 +1,211 @@
+# Functions
+
+Many Bash scripts are written in a strictly imperative style: execute one command, then execute another command, and so on.
+Sometimes you need to group together a sequence of commands that conceptually perform a single purpose.
+This is where _functions_ come in.
+
+## Defining a Function
+
+You define a function like this:
+
+```bash
+my_function () {
+    COMMANDS
+}
+```
+
+The empty set of parentheses simply denotes that you are defining a function.
+Nothing goes inside them.
+
+## Function Parameters
+
+Functions, once defined, act like any other command (builtin or not).
+Like any command, you can provide _arguments_ to your functions.
+Inside the functions, you access the arguments using the _positional parameters_, `$1`, `$2`, etc.
+(Recall, we learned about positional parameters in the [Variables][variables] concept.)
+
+~~~~exercism/advanced
+The special parameter `$0` is not changed inside a function; it is still the name of the executing script.
+The currently executing function can access its name with the `$FUNCNAME` variable.
+
+See [3.4.2 Special Parameters][special] in the manual.
+
+[special]: https://www.gnu.org/software/bash/manual/bash.html#Special-Parameters
+~~~~
+
+## Variables
+
+You can define variables inside a function.
+If you declare the variables with the `local` command, the _scope_ of the variable is limited to the current function (and to any functions called by it).
+Otherwise, the variable is placed in the _global scope_.
+
+Local variables can have the same name as a global variable.
+In that case, the local variable _temporarily_ overrides the global one, and the global value is restored when the function returns.
+
+```bash
+x=5
+
+myfunc () {
+    local x=100
+    echo "in my function, $x == 100"
+}
+
+echo "in the global scope, $x == 5"
+
+myfunc
+
+echo "back in the global scope, $x == 5"
+```
+
+This outputs
+
+```none
+in the global scope, 5 == 5
+in my function, 100 == 100
+back in the global scope, 5 == 5
+```
+
+Inside a function, you can access variables from the _caller_'s scope.
+That means you can use global variables, as well as local variables that were declared in the caller (or in some function that calls the caller).
+
+~~~~exercism/advanced
+Technically, "global" is not the right word to use.
+To expand a variable in a function, Bash will traverse up the call stack, as far as the global scope, to find a function where that variable name has been declared.
+
+This example is adapted from the [Shell Functions][man-funcs] section of the manual:
+
+```bash
+func1() {
+    local var='func1 local'
+    func2
+}
+
+func2() {
+    echo "In func2, var = $var"
+}
+
+var=global
+func1
+func2
+```
+
+The output is:
+
+```none
+In func2, var = func1 local
+In func2, var = global
+```
+
+Similarly, _assigning_ a value to a variable will assign it _in the scope where it was declared_.
+This "action at a distance" can create hard-to-follow code, as it is not always obvious where a variable was assigned a value.
+~~~~
+
+## Return Values
+
+A function, like any command, has an _exit status_.
+By default, the status of a function is the exit status of the _last command executed_.
+
+You can use the `return` command to return from a function with a specific exit status.
+
+```bash
+check_password () {
+    if [[ $1 == "secret" ]]; then
+        return 0
+    else
+        return 1
+    fi
+}
+
+read -sp "Enter your password: " pass
+
+if check_password "$pass"; then
+    echo "Correct!"
+else
+    echo "Wrong password."
+fi
+```
+
+Using `return` with no arguments returns the status of the last command executed.
+
+~~~~exercism/note
+Note that the `check_password` function can be simplified to:
+
+```bash
+check_password () { [[ $1 == "secret" ]]; }
+```
+
+1. The `[[...]]` conditional construct has an exit status: `0` for "true", `1` for "false.
+2. The `{...}` grouping construct must have either a newline or a semicolon before the ending brace.
+~~~~
+
+## Function Output
+
+The return status of a function is just a number.
+How can a function produce output?
+
+Your function can print to standard output.
+Use the familiar _command substitution_ to capture it:
+
+```bash
+d6 () { echo "$(( 1 + RANDOM % 6 ))"; }
+
+die=$( d6 )
+echo "You rolled a $die."
+```
+
+### Using Both the Output and the Status
+
+The exit status of a function is available to use even when you are capturing the output.
+
+```bash
+roll () {
+    local n=$1
+    if (( 4 <= n && n <= 20 )); then
+        echo "$(( 1 + RANDOM % n ))"  # exit status is 0
+    else
+        return 1
+    fi
+}
+
+read -p "How many faces does your die have? " faces
+if die=$( roll "$faces" ); then
+    echo "You rolled a $die."
+else
+    echo "I can't roll a die with $faces faces."
+fi
+```
+
+## Recursion
+
+Functions can call themselves recursively.
+By default, there is no limit to the depth of recursion.
+
+An example:
+
+```bash
+fibonacci() {
+    local n=$1
+    if (( n <= 1 )); then
+        echo "1"
+    else
+        local a=$(fibonacci "$(( n - 1 ))")
+        local b=$(fibonacci "$(( n - 2 ))")
+        echo "$(( a + b ))"
+    fi
+}
+
+for i in {1..10}; do fibonacci "$i"; done
+# => 1
+# => 2
+# => 3
+# => 5
+# => 8
+# => 13
+# => 21
+# => 34
+# => 55
+# => 89
+```
+
+[variables]: https://exercism.org/tracks/bash/concepts/variables
+[man-funcs]: https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions
--- a/concepts/functions/about.md
+++ b/concepts/functions/about.md
@@ -102,6 +102,12 @@ Similarly, _assigning_ a value to a variable will assign it _in the scope where 
 This "action at a distance" can create hard-to-follow code, as it is not always obvious where a variable was assigned a value.
 ~~~~
 
+~~~~exercism/advanced
+The call stack can be examined using [the `FUNCNAME` array variable][funcname].
+
+[funcname]: https://www.gnu.org/software/bash/manual/bash.html#index-FUNCNAME
+~~~~
+
 ## Return Values
 
 A function, like any command, has an _exit status_.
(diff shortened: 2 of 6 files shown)


concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
+# Functions
+
+Many Bash scripts are written in a strictly imperative style: execute one command, then execute another command, and so on.
+But often you'll need to have a group of commands that conceptually perform a single purpose.
+This is where _functions_ come in.
+
+## Defining a Function
+
+You declare a function is one of two ways.
```
kotp, line 9:
```suggestion
You declare a function in one of two ways.
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+The first is a "portable" style
+
+```bash
+funcname () { COMMANDS; }
```
kotp, line 13: Can we use `function_name` here?

It also shows what is,  I think, the style used for multiple names, while also using full words instead of abbreviations, so is helpful as documentation in its clarity as well.

Showing what I mean in a later suggestion...

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+This is the style that was created with the original Bourne shell.
+
+Alternately, you can use the `function` keyword
+
+```bash
+function funcname { COMMANDS; }
```
kotp, line 21:
```suggestion
function function_name { COMMANDS; }
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+Functions, once defined, act like any other command (builtin or not).
+Like any command, you can provide _arguments_ for your functions.
+Inside the functions, you access the arguments using the _positional parameters_, `$1`, `$2`, etc.
+(Recall, we learned about positional parameters in the [Variables][variables] concept.)
+
+~~~~exercism/advanced
+The special parameter `$0` is not changed inside a function; it is still the name of the executing script.
+The currently executing function can access its name with the `$FUNCNAME` variable.
+~~~~
```
kotp, line 36: We could give a link to the special parameters section manual page:

https://www.gnu.org/software/bash/manual/html_node/Special-Parameters.html

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+```
+
+Inside a function, you can access variables from the _caller_'s scope.
+That means you can use global variables, or local variables that are declared in some function that calls this one.
+
+~~~~exercism/advanced
+Technically, "global" is not the right word to use.
+Assignments to non-local variables will assign to the variable with that name that has been declared in some previous scope, up to the global scope.
+
+This example is taken from the bash manual
```
kotp, line 69: Which section of the man page?  I am not finding this in the man pages section 1.  Or a better question which bash manual?

Can we give the command or link to get to it?

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+else
+    echo "I can't roll a die with $faces faces."
+fi
+```
+
+## Recursion
+
+Functions can call themselves recursively.
+By default, there is no limit to the depth of recursion.
+
```
kotp, line 171:
```suggestion
```

kotp (commented):
Still in WIP/Draft state...

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+1. The `[[...]]` conditional construct has an exit status: `0` for "true", `1` for "false.
+2. The `{...}` grouping construct must have either a newline or a semicolon before the ending brace.
+~~~~
+
+## Function Output
+
+The return status of a function is just a number.
+How can a function produce output?
+
+Your function simply emits output on standard output.
```
IsaacG, line 135:
```suggestion
Your function can print to standard output.
Then, you can use command substitution to capture it. 
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+Use the familiar _command substitution_ to capture it:
+
+```bash
+d6 () { echo $(( 1 + RANDOM % 6 )); }
```
IsaacG, line 139: Always quote every expansion ;)

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+die=$( d6 )
+echo "You rolled a $die."
+```
+
+### Using Both the Output and the Status
+
+The exit status of a function is still available to use when you are capturing the output.
```
IsaacG, line 147:
```suggestion
The exit status of a function is available to use even when you are capturing the output.
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+read -p "How many faces does your die have? " faces
+if die=$( roll "$faces" ); then
+    echo "You rolled a $die."
+else
+    echo "I can't roll a die with $faces faces."
+fi
+```
+
+## Recursion
```
IsaacG, line 167: No examples? Fibonacci would be a classic.

concepts/functions/introduction.md
glennj: good point. Perhaps `my_function` since it doesn't _start_ with "function".

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+This is the style that was created with the original Bourne shell.
+
+Alternately, you can use the `function` keyword
+
+```bash
+function funcname { COMMANDS; }
+```
+
+There is no difference between the two styles.
```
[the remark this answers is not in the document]
glennj: I don't think so.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+
+The output is:
+
+```none
+In func2, var = func1 local
+In func2, var = global
+```
+~~~~
+
+## Return Value
```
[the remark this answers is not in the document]
glennj: a function can have multiple parameters and multiple local variables but only one return value.

concepts/functions/.meta/config.json
```diff
@@ -3,6 +3,8 @@
     "glennj"
   ],
   "contributors": [
+    "kotp",
+    "IsaacG"
   ],
   "blurb": "Functions in bash programs."
```
kotp, line 9:
```suggestion
  "blurb": "Functions in Bash programs."
```

Reason: This is the name of the language, so English says "capitalize first letter".  it also differentiates visually from the executable and the language itself.

concepts/functions/introduction.md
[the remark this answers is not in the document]
IsaacG: My personal preference, based on my time in the IRC channel, is for the portable version.

Mentioning two approaches without any suggestion in one direction or the other can leave someone new to the syntax a bit ... confused/adrift about which to use.

concepts/functions/introduction.md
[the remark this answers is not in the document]
IsaacG: A function call has one return value. A function can have many calls and many return values.

A function variable has one name and one value (at any given time).

It feels very odd to me for them to be inconsistent.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,206 @@
...
+read -sp "Enter your password: " pass
+
+if check_password "$pass"; then
+    echo "Correct!"
+else
+    echo "Wrong password."
+fi
+```
+
+Using `return` with no arguments returns a zero status.
```
IsaacG, line 134:
```suggestion
Using `return` with no arguments returns the status of the last command executed.
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,206 @@
...
+
+Functions can call themselves recursively.
+By default, there is no limit to the depth of recursion.
+
+An example:
+
+```bash
+factorial() {
+    local n=$1
+    if ((n <= 1)); then
```
IsaacG, line 194: The prior snippet using `((` space pads the `((` command.

```suggestion
    if (( n <= 1 )); then
```

concepts/functions/introduction.md
```diff
@@ -0,0 +1,206 @@
...
+    echo "You rolled a $die."
+else
+    echo "I can't roll a die with $faces faces."
+fi
+```
+
+## Recursion
+
+Functions can call themselves recursively.
+By default, there is no limit to the depth of recursion.
```
IsaacG, line 187: No recursion limit ... but factorials would overflow pretty quickly :smile:

IsaacG (approved):

glennj:
Regarding the 2 ways to define a function, I simply removed one of them. Frustratingly even the google shell style guide mixes the styles.

concepts/functions/about.md
```diff
@@ -0,0 +1,211 @@
...
+~~~~
+
+## Variables
+
+You can define variables inside a function.
+If you declare the variables with the `local` command, the _scope_ of the variable is limited to the current function (and to any functions called by it).
+Otherwise, the variable is placed in the _global scope_.
+
+Local variables can have the same name as a global variable.
+In that case, the local variable _temporarily_ overrides the global one, and the global value is restored when the function returns.
```
kotp, line 43: Can we explain this in a different way, where it is more clear what is happening.  Effectively, the description is accurate, but the global value is only overwritten, not in a state where it must be restored, it is not like it was "gone" or "delete".  Is this, maybe, a situation where the lookup table order matters, and when found we do not look for anything else with the same name?

The global really did not go anywhere, it is more like the name of the variable is found, and so there is no reason to continue looking, the next place to look being from the local scope to the global scope.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,173 @@
...
+~~~~
+
+## Variables
+
+You can define variables inside a function.
+If you declare the variables with the `local` command then the _scope_ of the variable is limited to the current function (and to any functions called from it).
+Otherwise, the variable is placed in the _global scope_.
+
+Local variables can have the same name as a global variable.
+In that case, the local variable _temporarily_ overrides the global one, and the global value is restored when the function returns.
```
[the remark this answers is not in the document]
kotp: Having some thoughts about this line as well.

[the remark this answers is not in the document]
kotp: Maybe it is covered well enough in lines +72 to +73

concepts/functions/about.md
kotp: As mentioned in the diff view, maybe it is explained well in  lines +72 to +73

kotp (approved):

concepts/functions/about.md
glennj: I am thinking about this, but I have a holiday cold so I can't real concentrate ATM

concepts/functions/introduction.md
```diff
@@ -40,7 +40,7 @@ If you declare the variables with the `local` command, the _scope_ of the variab
 Otherwise, the variable is placed in the _global scope_.
 
 Local variables can have the same name as a global variable.
-In that case, the local variable _temporarily_ overrides the global one, and the global value is restored when the function returns.
+In that case, the local variable takes precedence over the global variable.
```
kotp, line 43: Does it first use the local variable and then use the global variable?  To precede something means to come first, implying that the latter comes as well.  I think that was why the wording was used to not imply that something comes first, and that the global is not available at that scope.

At least that is how I understood it, and was personally glad to not see "precedence" used here. I can not speak for the author, though.

kotp: Is it perhaps "not available" or "not visible" due to the scope?

glennj: This is a hard concept to explain. The latest commit copies sentences directly from the manual.

kotp: I agree with that.

I sometimes do so graphically, and the shadowing idea follows that.

concepts/functions/about.md
```diff
@@ -0,0 +1,213 @@
...
+    echo "You rolled a $die."
+else
+    echo "I can't roll a die with $faces faces."
+fi
+```
+
+## Recursion
+
+Functions can call themselves recursively.
+By default, there is no limit to the depth of recursion.
```
kotp, line 183: "By default" ... this suggests that there is a setting or option where we can set a limit to recursion depth.

glennj: That's because there is:

> Functions may be recursive. The `FUNCNEST` variable may be used to limit the depth of the function call stack and restrict the number of function invocations. By default, no limit is placed on the number of recursive calls.

kotp: So a footnote link, here, maybe?  It will answer the implied question by providing a place where it may be answered for the reader?

Something that says "The manual has more information", if that makes sense in the eye of a student learning.

Or "see man bash and the FUNCNEST constant for more information."

concepts/functions/about.md
```diff
@@ -0,0 +1,245 @@
...
+4
+5
+environment: line 1: recur: maximum function nesting level exceeded (5)
+```
+
+[funcnest]: https://www.gnu.org/software/bash/manual/bash.html#index-FUNCNEST
+~~~~
+
+[variables]: https://exercism.org/tracks/bash/concepts/variables
+[man-funcs]: https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions
```
kotp, line 245: Should the funcnest reference link be with the other reference links?  Seems out of place.

concepts/functions/introduction.md
```diff
@@ -0,0 +1,245 @@
...
+4
+5
+environment: line 1: recur: maximum function nesting level exceeded (5)
+```
+
+[funcnest]: https://www.gnu.org/software/bash/manual/bash.html#index-FUNCNEST
+~~~~
+
+[variables]: https://exercism.org/tracks/bash/concepts/variables
+[man-funcs]: https://www.gnu.org/software/bash/manual/bash.html#Shell-Functions
```
kotp, line 245: Same here.

concepts/functions/about.md
glennj: The admonition blocks have to be self-contained

concepts/functions/about.md
IsaacG: Something like this?

```
Local variables can have the same name as a global variable.
Local and global refer to "scopes".
Each scope has its own set of variables.
When looking up a variable by name, bash checks the local variables first.
If it finds the variable in the local table, that's what it will use when reading and setting the variable.
Otherwise, it uses the global variables from the global table.
```

glennj: This has been updated in the meantime. How do you feel about the current text?

> Local variables can have the same name as a global variable. In that case, the local variable "shadows" the global variable. For instance, a local variable declared in a function hides a global variable of the same name: references and assignments refer to the local variable, leaving the global variable unmodified. When the function returns, the global variable is once again visible.


https://exercism.org/tracks/bash/concepts/functions#h-variables

kotp: "When the function returns"

What if the function is never called?  Is the "return" key here?  Should we express this in execution scope?  Perhaps it is fine.  We probably want to be careful to not suggest that a return is the trigger, but the environment that we are at while we are invoking statements.  Is "scope" a word that helps us here?

kotp: I read [this](https://github.com/exercism/bash/pull/715#discussion_r1921087855) before I read what was [here](https://github.com/exercism/bash/pull/715#discussion_r1920978336) and agree with the talk about scope.  The idea that the name of the variables are the same, but they are not the same things is probably going to help understand what effectively happens.

IsaacG: The current text seems reasonable to me!

glennj: This particular concept is really difficult to put into words.

Perl also has `local` that acts the same way. Eventually they invented `my`, a truly lexically-scoped qualifier.

IsaacG: Python has a `global` keyword to differentiate local from global ... as well as a `nonlocal` for that third scope :smile_cat:

PR #726, opened 2025-02-01
Concept: Arithmetic

glennj:


concepts/README.md
```diff
@@ -89,20 +93,21 @@ The [plan](http://forum.exercism.org/t/bash-syllabus-planning/11952)
...
-   - `exec` and redirections
    - command substitution
         - capturing stdout and stderr
         - capturing stdout and stderr **into separate variables**
+   - `exec` and redirections
+   - process substitutions
+
+## More Concepts
+
+- brace expansions and how it's different from patterns `/path/to/{foo,bar,baz}.txt`
```
IsaacG, line 110: We don't have globs in here :) People confuse globs and patterns, too.

concepts/README.md
```diff
@@ -115,3 +120,8 @@ x. `set` command and "strict mode"
    - `-opt` versus `+opt`
    - debugging with `-x` and `-v`
    - shellcheck
+
+   - https://mywiki.wooledge.org/BashFAQ/105
```
IsaacG, line 124: Hah! All the wooledge!

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+The arithmetic operators are listed (in order of precedence) in the manual: [Shell Arithmetic][arithmetic].
+
+Additionally, Bash only performs arithmetic in certain places (known as arithmetic "contexts").
+
+## Arithmetic Expansion
+
+Arithmetic expansion looks like parameter expansion, but with double parentheses.
+
+```bash
+$((expression))
```
IsaacG, line 15: Were we using spaces inside `(( ... ))` in the last doc?

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+Bash can do arithmetic but _only with integers_.
+If you need to do floating-point math, you'll need to call out to some external program (like [awk][awk] or [bc][bc]).
+
+The arithmetic operators are listed (in order of precedence) in the manual: [Shell Arithmetic][arithmetic].
+
+Additionally, Bash only performs arithmetic in certain places (known as arithmetic "contexts").
+
+## Arithmetic Expansion
+
+Arithmetic expansion looks like parameter expansion, but with double parentheses.
```
IsaacG, line 12: Did you already introduce command substitution? Arithmetic expansion looks more like that. I wouldn't really think arithmetic expansion looks like PE a ton.

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+
+```bash
+$((expression))
+```
+
+This expansion returns the result of evaluating the expression.
```
IsaacG, line 18:
```suggestion
This expansion is expanded to the result of evaluating the expression.
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+It is common to capture the result in a variable:
```
IsaacG, line 19:
```suggestion
It is commonly used to capture the result in a variable:
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+
+## Arithmetic Conditional
+
+The arithmetic conditional is performed within double parentheses _without_ the leading `$`:
+
+```bash
+((expression))
+```
+
+This does not return the result.
```
IsaacG, line 33:
```suggestion
This is not replaced with the result.
```

The exit status is sometimes referred to as the return code. This does produce a result which sets the return code.

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+However it does produce an _exit status_.
+
+* The exit status is 0 (success) if the expression's result is **non-zero**.
+* The exit status is 1 (failure) if the expression's result is **zero**.
+
+~~~~exercism/note
+Bash uses the same sense of true/false as the C language: zero is false, non-zero is true.
```
IsaacG, line 40:
```suggestion
Inside `((` Bash uses the same sense of true/false as the C language: zero is false, non-zero is true.
**This is the opposite of how Bash evaluates exit statuses!**
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+
+   ```bash
+   if ((x == y - 1)); then ...
+   while ((y > 0)); do ...
+   ```
+
+1. as a standalone command to modify a variable's value.
+
+   ```bash
+   ((++x))
```
IsaacG, line 55: You should pre-increment here then discuss post-increments. This is slightly confusing.

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+   echo "new value for x is $x"
+   ```
+
+   ~~~~exercism/caution
+   Be careful using arithmetic this way combined with `set -e` that the expression does not result in zero.
+   The exit status will be non-zero.
```
IsaacG, line 61:
```suggestion
   Be careful using arithmetic this way combined with `set -e`.
   If the expression evaluates to zero, the exit status will be non-zero and Bash will terminate.
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+   ```
+   ~~~~
+
+
+## Other Places With Arithmetic Context
+
+There are a few other places that are arithmetic contexts:
+
+* The [arithmetic for-loop][looping-for].
+* The index part of a numerically-indexed array.
```
IsaacG, line 81:
```suggestion
* The index part of a (non-associative) array.
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+   echo "shell exits on the previous line; you won't see this"
+   ```
+   ~~~~
+
+
+## Other Places With Arithmetic Context
+
+There are a few other places that are arithmetic contexts:
+
+* The [arithmetic for-loop][looping-for].
```
IsaacG, line 80: Should this be discussed/shown above?

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+
+  ~~~~exercism/note
+  It is the author's opinion that `declare -i` has limited usage.
+  It's main benefit is to allow "bare" arithmetic (without the arithmetic expression syntax).
+  This can cause confusion for readers of your code, where arithmetic expressions are evaluated in unexpected places.
+  ~~~~
+
+## Using Variables in an Arithmetic Expressions
+
+There is one aspect of Bash arithmetic that improves readability.
```
IsaacG, line 113:
```suggestion
Here is one aspect of arithmetic contexts that improves readability.
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+
+```bash
+height=10
+width=20
+area=$((height * width))
+```
+
+Did you see that the `$` dollar signs are missing from the variables in the arithmetic expression?
```
IsaacG, line 121:
```suggestion
Did you notice that the `$` dollar signs are missing from the variables in the arithmetic expression?
```

concepts/arithmetic/introduction.md
```diff
@@ -0,0 +1,143 @@
...
+Bash lets you do that.
+It makes expressions much easier to read.
+
+The manual says it this way:
```
IsaacG, line 125:
```suggestion
The manual says,
```

concepts/arithmetic/introduction.md
glennj: I'll make a note that spaces are allowed.

concepts/README.md
glennj: There's a whole section about it in Quoting. Do we need more?

concepts/arithmetic/introduction.md
glennj: I want to emphasis the "numeric" aspect.

concepts/arithmetic/introduction.md
glennj: I don't think so. `for` requires 3 expressions separated by semicolons, which is specialized synax.

concepts/README.md
IsaacG: Ah. Nope. [There is a glob section](https://github.com/exercism/bash/blob/main/concepts/quoting/introduction.md#filename-expansion). It's just named "filename expansion" :)

concepts/arithmetic/introduction.md
IsaacG: I meant more that we should use a consistent style across docs.

concepts/arithmetic/introduction.md
IsaacG: If you make an associative array and use numbers as keys, is it numerically indexed?

glennj: Nope.
```
x=5
declare -A a
a[x + 1]=10
declare -p a
# => declare -A a=(["x + 1"]="10" )

```

glennj: But beyond the details, that's how I denote the 2 kinds of arrays. And that's the language I use in the Arrays doc.

concepts/arithmetic/introduction.md
glennj: Yes, in the Functions doc.

What's your preference? Spaces or no spaces

glennj:
Another thing I just realized. Considering the concept flow (http://forum.exercism.org/t/bash-syllabus-flow/15038/6) I have links from here into the sibling concept Looping.

Problematic?

concepts/arithmetic/introduction.md
IsaacG: I prefer spaces but I'm not strongly opinionated here.

concepts/arithmetic/introduction.md
IsaacG:
```bash
declare -A a
a[5]=10
a[6]=12
a[7]=14
declare -p a
# declare -A a=([7]="14" [6]="12" [5]="10" )
```

Aside from the unbalanced spacing, that looks like a numerically indexed array ;) I think the language/term can be ambiguous. But if you prefer that language, that works for me.

concepts/arithmetic/introduction.md
IsaacG: Sounds good. Having it in the looping doc instead seems reasonable, too.

IsaacG:
> Another thing I just realized. Considering the concept flow (http://forum.exercism.org/t/bash-syllabus-flow/15038/6) I have links from here into the sibling concept Looping.
> 
> Problematic?

They both unlock at the same time so I think this is fine.

IsaacG (approved):

kotp (approved):

concepts/arithmetic/introduction.md
glennj: This would be a good place for a `(For more info on arrays, see the [Arrays] concept doc)` link.

PR #729, opened 2025-02-21
Arrays concept part 2

glennj:


concepts/more-arrays/introduction.md
```diff
@@ -0,0 +1,197 @@
+# More about Arrays
+
+We were introduced to arrays in the [Arrays][arrays] chapter.
+This document will show more ways to use arrays.
+
+## Concatenating the Elements of an Array into a Single String
+
+In the previous Arrays chapter, you saw `"${myarray[@]}"`, with the `@` index, to expand the array into the individual elements.
```
IsaacG, line 8:
```suggestion
In the previous Arrays chapter, you saw `"${myarray[@]}"`, with the `@` index, used to expand the array into the individual elements.
```

concepts/more-arrays/introduction.md
```diff
@@ -0,0 +1,197 @@
...
+But sometimes you want to join all the elements into a single string.
+For this, use the `*` index:
+
+```bash
+echo "${myarray[*]}"
+```
+
+You are required to enclose the expansion in double quotes.
```
IsaacG, line 16: There's no actual "requirement" as much as a "should" to avoid glob expansion.

concepts/more-arrays/introduction.md
```diff
@@ -0,0 +1,197 @@
...
+    shift
+    local elements=("$@")
+    echo "${elements[*]}"
+}
+
+join ":" "${myarray[@]}"     # note, the "@" index
+# => "one:two:three:four"
+```
+
+Localizing `IFS` in the function means we don't have to save the old value and restore it back to it's previous value in the global scope.
```
IsaacG, line 53: FWIW a subshell is a (simpler?) alternative.

```bash
array=(one two three)
( IFS=,; echo "${array[*]}"; )
```

concepts/more-arrays/introduction.md
```diff
@@ -0,0 +1,197 @@
...
+* They are concatenated into a single string with `"$*"`
+* The number of parameters is `$#`
+
+Use the `set` command to assign values to them:
+
+```sh
+set -- one two three
+set -- "$@" four
+
+for item in "$@"; do
```
IsaacG, line 189: FWIW you can also omit the `in "$@"`

concepts/more-arrays/introduction.md
glennj: The requirement is that the concatenate behaviour only applies when the expansion is quoted. Otherwise you get word splitting. 

In the manual:

> If the word is double-quoted, `${name[*]}` expands to a single word with the value of each array member separated by the first character of the `IFS` variable,

concepts/more-arrays/introduction.md
glennj: Yes, but let's be explicit

concepts/more-arrays/introduction.md
glennj: Will add it

IsaacG (approved):

PR #740, opened 2025-05-27
Sort practice exercises by difficulty then slug.

IsaacG:
```
jq '.exercises.practice = (.exercises.practice|sort_by(.difficulty, .slug))' config.json > t && mv t config.json
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -46,6 +46,17 @@
         ]
       },
       {
+        "slug": "reverse-string",
+        "name": "Reverse String",
+        "uuid": "2b6d4faa-1909-422a-8eac-1682be3f1b9e",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 1,
+        "topics": [
+          "strings"
+        ]
+      },
+      {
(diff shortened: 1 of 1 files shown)


glennj (approved):

IsaacG:
> Approving, but why? Why not any other organization of this file?

The order of the exercises in this file dictate the order they show up on the website. Nothing else in this file has ordering that shows up on the website.

> What prompted the change?

A PR in another repo where I noticed the exercises were not in difficulty order.

kotp:
> > Approving, but why? Why not any other organization of this file?
> 
> The order of the exercises in this file dictate the order they show up on the website. Nothing else in this file has ordering that shows up on the website.
> 
> > What prompted the change?
> 
> A PR in another repo where I noticed the exercises were not in difficulty order.

Great answer.  That could be in the body of the commit for "Story telling time" when we look at the git logs.  But perhaps it is too late, as it appears it is in the main branch now, and that should not be ammended.

PR #770, opened 2026-05-05
[crypto-square] Remove `perl` usage from example solution

IsaacG:


IsaacG:
I was trying to change the test runner from Ubuntu to Alpine but this example solution uses `perl`. I don't think we want to provide `perl` as part of the tooling/test runner image. Removing it from the example.

IsaacG:
+cc @kotp and/or @glennj  for merging

kotp (approved):

PR #771, opened 2026-05-05
[meetup] Remove `perl` usage from example solution

IsaacG:


IsaacG:
I was trying to change the test runner from Ubuntu to Alpine but this example solution uses `perl`. I don't think we want to provide `perl` as part of the tooling/test runner image. Removing it from the example.

See also #770 

+cc @kotp

kotp (approved):

glennj (approved):

PR #772, opened 2026-05-05
[yacht] Remove dependency on GNU paste from example solution

IsaacG:


IsaacG:
Alpine ships with busybox which includes BusyBox paste. The BusyBox paste does not support `paste -d ''`. `tr` works fine here.

kotp (approved):

PR #775, opened 2026-05-31
Align instruction exemples with `assert_output` expected in `secret handshake` exercice

resu-xuniL:
In this exercice, there's a difference between the output expected (`wink,double blink`) and the exemples given in the `instructions.md` file (`wink, jump`) -> **extra space after comma**.

This PR is to align them.

I choose to edit the `instructions.md` file so the tests have not to be run again.

If the intent was that the outputs should have that extra space after comma, the `secret_handshake.bats` file needs to be modified but in that case, all the tests must be run again as the `assert_output` would no longer be the same.

[no important files changed]
---

Reviewer Resources:

### DIFF
--- a/exercises/practice/secret-handshake/.docs/instructions.md
+++ b/exercises/practice/secret-handshake/.docs/instructions.md
@@ -26,7 +26,7 @@ Let's use the number `9` as an example:
 That was the last digit, so the final code is:
 
 ```plaintext
-wink, jump
+wink,jump
 ```
 
 Given the number 26, which is `11010` in binary, we get the following actions:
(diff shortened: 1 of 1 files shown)


IsaacG:
https://exercism.org/docs/community/being-a-good-community-member/suggesting-exercise-improvements

That markdown file is synced from an upstream repo. We generally do not accept manual changes to that file.

IsaacG:
Given the limitations of using STDOUT for this track, the tests tend to be a touch wonky. The student is expected to use the tests to discover the exact requirements. I'm not sure we want to change the tests here, either. At most, maybe an append file for the instructions. I'd love to get the thoughts of the other maintainers.

glennj:
I'm inclined to leave it as-is. Students may see the tests fail, spot the difference, and update their code.

PR #777, opened 2026-06-06
Fix sorted exercise workload URL

IsaacG:


.github/workflows/sorted.yml
```diff
@@ -11,6 +11,6 @@ permissions:
 jobs:
   call-gha-workflow:
     name: Check
-    uses: IsaacG/github-actions/.github/workflows/sorted.yml@main
+    uses: exercism/github-actions/.github/workflows/sorted.yml@main
```
glennj, line 14: Use a SHA instead of "main"?

glennj (approved):

.github/workflows/sorted.yml
IsaacG:
```
exercism » grep -h uses:.exercism */.github/workflows/*.yml | cut -d @ -f2 | sort | uniq -c
      1 backup-repo
      2 f4a6aa5acc07742989788e70fd89cdc0980f0d1e
   1340 main
      1 v1.5.2
      3 v2.1
```

We SHA-pin external workflows ... but mainly use `main` for Exercism workflows. Given our direct control, I think that's fine and reasonable?

glennj (approved):

PR #779, opened 2026-06-12
Add `baffling-birthdays` exercise

resu-xuniL:
Add baffling-birthdays exercice to track.
I change difficulty from 1 to 4.

Hope everything is correct.

 float with bash !
 I increase the tolerance for the exercice with 10 and 23 people because, after multiple runs (many time consuming), sometimes one or the other fail when checking only respectively `11|12|13` and  `50|51|52`.
 or might be  another way to do it more precisely ?

The other option is to increase the number of runs, but this is very time consuming in bash! (set now at 1000)
maybe I can try to parallelize the computings? or creating a file would help?
 

I got these result while debbuging  : 

``` ✗ estimated probability of at least one shared birthday -> among ten people
   (from function `assert_output' in file bats-extra.bash, line 394,
    in test file baffling_birthdays.bats, line 162)
     `assert_output $expect' failed
   
   -- output differs --
   expected : 11.694818
   actual   : 12.70
   --
   
 ✗ estimated probability of at least one shared birthday -> among twenty-three people
   (from function `assert_output' in file bats-extra.bash, line 394,
    in test file baffling_birthdays.bats, line 170)
     `assert_output $expect' failed
   
   -- output differs --
   expected : 50.729723
   actual   : 51.10
   --
   
 ✗ estimated probability of at least one shared birthday -> among seventy people
   (from function `assert_output' in file bats-extra.bash, line 394,
    in test file baffling_birthdays.bats, line 178)
     `assert_output $expect' failed
   
   -- output differs --
   expected : 99.915958
   actual   : 99.90
   --```
---

Reviewer Resources:

### DIFF
--- a/config.json
+++ b/config.json
@@ -656,6 +656,14 @@
         ]
       },
       {
+        "slug": "baffling-birthdays",
+        "name": "Baffling Birthdays",
+        "uuid": "c3fbce6a-61ce-439c-8c67-fad93561326f",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4
+      },
+      {
         "slug": "binary-search",
         "name": "Binary Search",
         "uuid": "47ff0fd8-178a-47fd-ab53-382ae1bb4f03",
--- a/config.json
+++ b/config.json
@@ -661,7 +661,10 @@
         "uuid": "c3fbce6a-61ce-439c-8c67-fad93561326f",
         "practices": [],
         "prerequisites": [],
-        "difficulty": 4
+        "difficulty": 4,
+        "topics": [
+          "dates"
+        ]
       },
       {
         "slug": "binary-search",
--- a/exercises/practice/baffling-birthdays/.meta/example.sh
+++ b/exercises/practice/baffling-birthdays/.meta/example.sh
@@ -1,19 +1,21 @@
 #!/usr/bin/env bash
 
+current_year=$(date -d "now" +%Y)
+
 main() {
     local function=$1; shift
 
-    $function "$@"
+    "$function" "$@"
 }
 
-sharedBirthday () {
+shared_birthday () {
     local -a birthdates=( "$@" )
     local -A birthdays
 
     for birthdate in "${birthdates[@]}"; do
         birthday=${birthdate: -5}
 
-        (( birthdays[$birthday] ++ ))
+        (( birthdays[$birthday]++ ))
     done
 
     if (( ${#birthdays[@]} == ${#birthdates[@]} )); then
@@ -23,54 +25,51 @@ sharedBirthday () {
     fi
 }
 
-randomBirthdates () {
+random_int() {
+    echo $(( RANDOM % $1 ))
+}
+
+random_birthdates () {
     local -i number=$1
-    local -i actual_year random_year random_month random_day
+    local -i random_year random_month random_day
     local -a birthdates
 
-    actual_year=$(date -d "now" +%Y)
-
-    for ((i = 0; i < number; i++)); do
-        random_year=$(shuf -n 1 -i 1900-${actual_year})
-        random_month=$(shuf -n 1 -i 1-12)
+    for (( i = 0; i < number; i++ )); do
+        random_year=$(( 1900 + $(random_int $(( current_year - 1900 + 1 ))) ))
+        random_month=$(( 1 + $(random_int 12) ))
 
         if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
             (( random_year ++ ))
         fi
 
-        if (( random_month == 2 )); then
-            random_day=$(shuf -n 1 -i 1-28)
-        elif [[ $random_month =~ ^(4|6|9|11)$ ]]; then
-            random_day=$(shuf -n 1 -i 1-30)
-        else
-            random_day=$(shuf -n 1 -i 1-31)
-        fi
+        case "$random_month" in
+                   2) random_day=$(( 1 + $(random_int 28) )) ;;
+            4|6|9|11) random_day=$(( 1 + $(random_int 30) )) ;;
+                   *) random_day=$(( 1 + $(random_int 31) )) ;;
+        esac
 
-        birthdates+=( "$(date -d "${random_year}-${random_month}-${random_day}" +%F)" )
+        birthdates+=( "$(printf '%d-%02d-%02d' "$random_year" "$random_month" "$random_day")" )
     done
 
     echo "${birthdates[@]}"
 }
 
-estimatedProbabilityOfSharedBirthday () {
+estimated_probability_of_share_birthday () {
     local -i group_size=$1
-    # the more you increase the 'runs' value, 
+    # the more you increase the 'runs' value,
     # the more the result is approching the expected one
     # but the more you increase the computing time!
-    local -i runs=1000
+    local -i runs=2000
     local -i count=0
-    local result
 
-
-    for ((i = 0; i <= runs; i++)); do
-        if [[ $(sharedBirthday $(randomBirthdates $group_size)) == "true" ]]; then
-            (( ++ count ))
+    for (( i = 0; i <= runs; i++ )); do
+        read -ra birthdates < <(random_birthdates "$group_size")
+        if [[ "$(shared_birthday "${birthdates[@]}")" == "true" ]]; then
+            (( count++ ))
         fi
     done
 
-    result="$(LC_NUMERIC=C printf "%.2f\n" "$(echo "scale=2; ($count * 100) / $runs" | bc)")"
-
-    echo "$result"
+    echo $(( count * 100 / runs ))
 }
 
 main "$@"
--- a/exercises/practice/baffling-birthdays/.meta/example.sh
+++ b/exercises/practice/baffling-birthdays/.meta/example.sh
@@ -26,7 +26,7 @@ shared_birthday () {
 }
 
 random_int() {
-    echo $(( RANDOM % $1 ))
+    echo "$(( RANDOM % $1 ))"
 }
 
 random_birthdates () {
@@ -39,7 +39,7 @@ random_birthdates () {
         random_month=$(( 1 + $(random_int 12) ))
 
         if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
-            (( random_year ++ ))
+            (( random_year++ ))
         fi
 
         case "$random_month" in
@@ -64,12 +64,12 @@ estimated_probability_of_share_birthday () {
 
     for (( i = 0; i <= runs; i++ )); do
         read -ra birthdates < <(random_birthdates "$group_size")
-        if [[ "$(shared_birthday "${birthdates[@]}")" == "true" ]]; then
+        if "$(shared_birthday "${birthdates[@]}")"; then
             (( count++ ))
         fi
     done
 
-    echo $(( count * 100 / runs ))
+    echo "$(( count * 100 / runs ))"
 }
 
 main "$@"
--- a/exercises/practice/baffling-birthdays/.meta/example.sh
+++ b/exercises/practice/baffling-birthdays/.meta/example.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+
+main() {
+    local function=$1; shift
+
+    $function "$@"
+}
+
+sharedBirthday () {
+    local -a birthdates=( "$@" )
+    local -A birthdays
+
+    for birthdate in "${birthdates[@]}"; do
+        birthday=${birthdate: -5}
+
+        (( birthdays[$birthday] ++ ))
+    done
+
+    if (( ${#birthdays[@]} == ${#birthdates[@]} )); then
+        echo "false"
+    else
+        echo "true"
+    fi
+}
+
+randomBirthdates () {
+    local -i number=$1
+    local -i actual_year random_year random_month random_day
+    local -a birthdates
+
+    actual_year=$(date -d "now" +%Y)
+
+    for ((i = 0; i < number; i++)); do
+        random_year=$(shuf -n 1 -i 1900-${actual_year})
+        random_month=$(shuf -n 1 -i 1-12)
+
+        if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
+            (( random_year ++ ))
+        fi
+
+        if (( random_month == 2 )); then
+            random_day=$(shuf -n 1 -i 1-28)
+        elif [[ $random_month =~ ^(4|6|9|11)$ ]]; then
+            random_day=$(shuf -n 1 -i 1-30)
+        else
+            random_day=$(shuf -n 1 -i 1-31)
+        fi
+
+        birthdates+=( "$(date -d "${random_year}-${random_month}-${random_day}" +%F)" )
+    done
+
+    echo "${birthdates[@]}"
+}
+
+estimatedProbabilityOfSharedBirthday () {
+    local -i group_size=$1
+    # the more you increase the 'runs' value, 
+    # the more the result is approching the expected one
+    # but the more you increase the computing time!
+    local -i runs=1000
+    local -i count=0
+    local result
+
+
+    for ((i = 0; i <= runs; i++)); do
+        if [[ $(sharedBirthday $(randomBirthdates $group_size)) == "true" ]]; then
+            (( ++ count ))
+        fi
+    done
+
+    result="$(LC_NUMERIC=C printf "%.2f\n" "$(echo "scale=2; ($count * 100) / $runs" | bc)")"
+
+    echo "$result"
+}
+
+main "$@"
--- a/exercises/practice/baffling-birthdays/.meta/example.sh
+++ b/exercises/practice/baffling-birthdays/.meta/example.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-current_year=$(date -d "now" +%Y)
+current_year=$(date -d 'now' +%Y)
 
 main() {
     local function=$1; shift
@@ -8,7 +8,7 @@ main() {
     "$function" "$@"
 }
 
-shared_birthday () {
+shared_birthday() {
     local -a birthdates=( "$@" )
     local -A birthdays
 
@@ -25,48 +25,83 @@ shared_birthday () {
     fi
 }
 
-random_int() {
-    echo "$(( RANDOM % $1 ))"
+# Sets global _random_birthday; avoids subshell fork on each call.
+_make_random_birthdate() {
+    local -i random_year random_month random_day
+
+    random_year=$(( 1900 + RANDOM % (current_year - 1900 + 1) ))
+
+    # skip leap years
+    if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
+        (( random_year++ ))
+    fi
+
+    random_month=$(( 1 + RANDOM % 12 ))
+
+    case "$random_month" in
+               2) random_day=$(( 1 + RANDOM % 28 )) ;;
+        4|6|9|11) random_day=$(( 1 + RANDOM % 30 )) ;;
+               *) random_day=$(( 1 + RANDOM % 31 )) ;;
+    esac
+
+    # # Zero-pad month and day without a printf subshell
+    # local mm dd
+    # (( random_month < 10 )) && mm="0${random_month}" || mm="${random_month}"
+    # (( random_day   < 10 )) && dd="0${random_day}"   || dd="${random_day}"
+
+    # _random_birthday="${random_year}-${mm}-${dd}"
+    printf -v _random_birthday "%04d-%02d-%02d" "$random_year" "$random_month" "$random_day"
 }
 
-random_birthdates () {
+random_birthdates() {
     local -i number=$1
-    local -i random_year random_month random_day
     local -a birthdates
 
     for (( i = 0; i < number; i++ )); do
-        random_year=$(( 1900 + $(random_int $(( current_year - 1900 + 1 ))) ))
-        random_month=$(( 1 + $(random_int 12) ))
-
-        if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
-            (( random_year++ ))
-        fi
-
-        case "$random_month" in
-                   2) random_day=$(( 1 + $(random_int 28) )) ;;
-            4|6|9|11) random_day=$(( 1 + $(random_int 30) )) ;;
-                   *) random_day=$(( 1 + $(random_int 31) )) ;;
-        esac
-
-        birthdates+=( "$(printf '%d-%02d-%02d' "$random_year" "$random_month" "$random_day")" )
+        _make_random_birthdate
+        birthdates+=( "$_random_birthday" )
     done
 
     echo "${birthdates[@]}"
 }
 
-estimated_probability_of_share_birthday () {
+# Sets global _has_shared (1=true, 0=false); avoids subshell fork on each call.
+_check_shared_birthday() {
+    local -A birthdays
+    local birthdate birthday
+
+    for birthdate in "$@"; do
+        birthday=${birthdate: -5}
+        (( birthdays[$birthday]++ ))
+    done
+
+    if (( ${#birthdays[@]} == $# )); then
+        _has_shared=0
+    else
+        _has_shared=1
+    fi
+}
+
+estimated_probability_of_share_birthday() {
     local -i group_size=$1
     # the more you increase the 'runs' value,
-    # the more the result is approching the expected one
+    # the more the result is approaching the expected one
     # but the more you increase the computing time!
     local -i runs=2000
     local -i count=0
+    local -a birthdates
 
     for (( i = 0; i <= runs; i++ )); do
-        read -ra birthdates < <(random_birthdates "$group_size")
-        if "$(shared_birthday "${birthdates[@]}")"; then
-            (( count++ ))
-        fi
+        # Generate birthdates inline - no subshell fork
+        birthdates=()
+        for (( j = 0; j < group_size; j++ )); do
+            _make_random_birthdate
+            birthdates+=( "$_random_birthday" )
+        done
+
+        # Check for shared birthday inline - no subshell fork
+        _check_shared_birthday "${birthdates[@]}"
+        (( count += _has_shared ))
     done
 
     echo "$(( count * 100 / runs ))"
--- a/exercises/practice/baffling-birthdays/baffling_birthdays.bats
+++ b/exercises/practice/baffling-birthdays/baffling_birthdays.bats
@@ -166,7 +166,7 @@ load bats-extra
     # expect="50.729723" # this is the exercice ref.
     run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 23
     assert_success
-    [[ "$output" =~ ^(49\.|50\.|51\.|52\.|53\.) ]]
+    [[ "$output" =~ ^(48\.|49\.|50\.|51\.|52\.|53\.) ]]
 }
 
 @test "estimated probability of at least one shared birthday -> among seventy people" {
--- a/exercises/practice/baffling-birthdays/baffling_birthdays.bats
+++ b/exercises/practice/baffling-birthdays/baffling_birthdays.bats
@@ -1,0 +1,178 @@
+#!/usr/bin/env bats
+load bats-extra
+
+@test "shared birthday -> one birthdate" {
+    #[[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("2000-01-01")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "false"
+}
+
+@test "shared birthday -> two birthdates with same year, month, and day" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("2000-01-01" "2000-01-01")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "shared birthday -> two birthdates with same year and month, but different day" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("2012-05-09" "2012-05-17")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "false"
+}
+
+@test "shared birthday -> two birthdates with same month and day, but different year" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("1999-10-23" "1988-10-23")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "shared birthday -> two birthdates with same year, but different month and day" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("2007-12-19" "2007-04-27")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "false"
+}
+
+@test "shared birthday -> two birthdates with different year, month, and day" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=("1997-08-04" "1963-11-23")
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "false"
+}
+
+@test "shared birthday -> multiple birthdates without shared birthday" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=(
+        "1966-07-29"
+        "1977-02-12"
+        "2001-12-25"
+        "1980-11-10"
+    )
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "false"
+}
+
+@test "shared birthday -> multiple birthdates with one shared birthday" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=(
+        "1966-07-29"
+        "1977-02-12"
+        "2001-07-29"
+        "1980-11-10"
+    )
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "shared birthday -> multiple birthdates with more than one shared birthday" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    birthdates=(
+        "1966-07-29"
+        "1977-02-12"
+        "2001-12-25"
+        "1980-07-29"
+        "2019-02-12"
+    )
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "random birthdates -> generate requested number of birthdates" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=$(shuf -n 1 -i 100-1000)
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    assert_equal ${#output_dates[@]} $generate
+}
+
+@test "random birthdates -> years are not leap years" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=$(shuf -n 1 -i 100-1000)
+    seen_leap=0
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        year=${date:0:4}
+        if (( year % 400 == 0 || (year % 4 == 0 && year % 100 != 0) )); then
+            (( ++ seen_leap ))
+        fi
+    done
+    assert_equal $seen_leap 0
+}
+
+@test "random birthdates -> months are random" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    declare -A seen_month
+    generate=$(shuf -n 1 -i 100-1000)
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        month=${date:5:2}
+
+        seen_month[$month]=true
+    done
+    assert_equal ${#seen_month[@]} 12
+}
+
+@test "random birthdates -> days are random" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    declare -A seen_day
+    generate=$(shuf -n 1 -i 100-1000)
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        day=${date:8}
+
+        seen_day[$day]=true
+    done
+    assert_equal ${#seen_day[@]} 31
+}
+
+# "The expected probability values should be compared using some tolerance to allow for small deviations."
+@test "estimated probability of at least one shared birthday -> for one person" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="0.0" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 1
+    assert_success
+   [[ "$output" =~ ^(0\.) ]]
+}
+
+@test "estimated probability of at least one shared birthday -> among ten people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="11.694818" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 10
+    assert_success
+    [[ "$output" =~ ^(10\.|11\.|12\.|13\.|14\.) ]]
+}
+
+@test "estimated probability of at least one shared birthday -> among twenty-three people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="50.729723" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 23
+    assert_success
+    [[ "$output" =~ ^(49\.|50\.|51\.|52\.|53\.) ]]
+}
+
+@test "estimated probability of at least one shared birthday -> among seventy people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="99.915958" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 70
+    assert_success
+    [[ "$output" =~ ^(98\.|99\.|100\.) ]]
+}
--- a/exercises/practice/baffling-birthdays/baffling_birthdays.bats
+++ b/exercises/practice/baffling-birthdays/baffling_birthdays.bats
@@ -158,7 +158,7 @@ load bats-extra
     # The expected probability is 11.694818
     run bash baffling_birthdays.sh estimated_probability_of_share_birthday 10
     assert_success
-    assert_between 10 13
+    assert_between 9 13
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: exercises/practice/baffling-birthdays/.meta/example.sh [hunk 1]
```diff
@@ -0,0 +1,76 @@
...
+
+    actual_year=$(date -d "now" +%Y)
+
+    for ((i = 0; i < number; i++)); do
+        random_year=$(shuf -n 1 -i 1900-${actual_year})
+        random_month=$(shuf -n 1 -i 1-12)
+
+        if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
+            (( random_year ++ ))
+        fi
```
IsaacG: Does the year need to be random?

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,76 @@
...
+
+    for ((i = 0; i <= runs; i++)); do
+        if [[ $(sharedBirthday $(randomBirthdates $group_size)) == "true" ]]; then
+            (( ++ count ))
+        fi
+    done
+
+    result="$(LC_NUMERIC=C printf "%.2f\n" "$(echo "scale=2; ($count * 100) / $runs" | bc)")"
+
+    echo "$result"
```
IsaacG: Don't capture output just to print the same thing back out.

```suggestion
    bc <<< "scale=2; $count * 100 / $runs"
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+        "2019-02-12"
+    )
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "random birthdates -> generate requested number of birthdates" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=$(shuf -n 1 -i 100-1000)
```
IsaacG: Do we need randomness in test inputs? Test inputs should generally be stable.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    run bash baffling_birthdays.sh randomBirthdates $generate
```
IsaacG:
```suggestion
    run bash baffling_birthdays.sh randomBirthdates "${generate}"
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_success
+    read -ra output_dates <<< "$output"
+    assert_equal ${#output_dates[@]} $generate
```
IsaacG:
```suggestion
    assert_equal "${#output_dates[@]}" "${generate}"
```

Always quote expansions.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        year=${date:0:4}
+        if (( year % 400 == 0 || (year % 4 == 0 && year % 100 != 0) )); then
+            (( ++ seen_leap ))
+        fi
+    done
+    assert_equal $seen_leap 0
```
IsaacG:
```suggestion
    assert_equal "${seen_leap}" 0
```

IsaacG (changes requested):

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: I guess it should be... shouldn't it?

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: I guess so (in `canonical-data.json`:)
```
          "uuid": "70b38cea-d234-4697-b146-7d130cd4ee12",
          "description": "generate requested number of birthdates",
          "scenarios": ["random"],
          "property": "randomBirthdates",
          "input": {},
          "expected": "length == groupsize"
```
Or I misunderstood?

exercises/practice/baffling-birthdays/.meta/example.sh
IsaacG: The date is supposed to be random. The year doesn't matter. You could simplify this a whole lot by just using a fixed year.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
IsaacG: A fixed group size should be fine here. Unit tests typically use fixed inputs. Randomizing inputs makes things like benchmarking annoying.

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,76 @@
...
+}
+
+sharedBirthday () {
+    local -a birthdates=( "$@" )
+    local -A birthdays
+
+    for birthdate in "${birthdates[@]}"; do
+        birthday=${birthdate: -5}
+
+        (( birthdays[$birthday] ++ ))
```
IsaacG: The `++` operator is usually not space padded.


```suggestion
        (( birthdays[$birthday]++ ))
```

resu-xuniL:
`23 persons` test failed again...
I'll work on it. Any suggestions are welcome! ^^
it passed just before the this test ...

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: in `canonical-data.json`:
```
          "uuid": "d9d5b7d3-5fea-4752-b9c1-3fcd176d1b03",
          "description": "years are not leap years",
          "scenarios": ["random"],
          "property": "randomBirthdates",
          "input": {},
          "expected": {
            "years": {
              "leapYear": false
            }
```

resu-xuniL: i thought that if the year is checked not to be a leap year, this would mean the year sould be random too.
am I wrong?

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: I'll change this

exercises/practice/baffling-birthdays/.meta/example.sh
IsaacG: The year cannot be a leap year. Nothing checks (hence, requires) that the year be random. The only year requirement is that the year not be a leap year.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_success
+	assert_equal "$output" 0
+}
+
+@test "estimated probability of at least one shared birthday -> among ten people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="11.694818" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 10
+    assert_success
+    [[ $output =~ ^(10\.|11\.|12\.|13\.|14\.) ]]
```
IsaacG: If you strip `.*`  you can do a simple number comparison.

IsaacG:
> Geez... `23 persons` test failed again... I'll work on it tomorrow. Any suggestions are welcome! ^^ it passed just before the this test ...

You can try doubling your runs from 1k to 2k. You can try widening the acceptable probability range.

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: I like the fact that the script itself should find no leap year. If it is already set, there is no challenge I think.
Plus, this is more realistic than everyone being born in the same year.

Shall I modify this behavior anyway?

resu-xuniL:
OK, it passes 6 times in a row (with this one) so far...
Thank you for your help!
(with runs at 1500)

exercises/practice/baffling-birthdays/.meta/example.sh
IsaacG: Eh. This is fine. So long as the example runs, it doesn't matter a ton. The year selection (1900-now) is also somewhat arbitrary but it's fine.

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,72 @@
...
+
+    for ((i = 0; i < number; i++)); do
+        random_year=$(shuf -n 1 -i 1900-${actual_year})
+        random_month=$(shuf -n 1 -i 1-12)
+
+        if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
+            (( random_year ++ ))
+        fi
+
+        if (( random_month == 2 )); then
```
IsaacG: A `case` would work really well here.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    )
+    run bash baffling_birthdays.sh sharedBirthday "${birthdates[@]}"
+    assert_success
+    assert_output "true"
+}
+
+@test "random birthdates -> generate requested number of birthdates" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=500
+    run bash baffling_birthdays.sh randomBirthdates $generate
```
IsaacG: Always quote every expansions.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=500
+    seen_leap=0
+    run bash baffling_birthdays.sh randomBirthdates $generate
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        year=${date:0:4}
+        if (( year % 400 == 0 || (year % 4 == 0 && year % 100 != 0) )); then
+            (( ++ seen_leap ))
```
IsaacG:
```suggestion
            (( seen_leap++ ))
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_success
+    read -ra output_dates <<< "$output"
+    assert_equal ${#output_dates[@]} $generate
+}
+
+@test "random birthdates -> years are not leap years" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=500
+    seen_leap=0
+    run bash baffling_birthdays.sh randomBirthdates $generate
```
IsaacG: Expansions should be quoted.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_success
+    assert_equal "$output" 0
+}
+
+@test "estimated probability of at least one shared birthday -> among ten people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="11.694818" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimatedProbabilityOfSharedBirthday 10
+    assert_success
+    (( output >= 10 || output <= 12 ))
```
IsaacG: Do you mean `||` or `&&` here?

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: OK then

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: Need I always quote them if they are integers ?
I use shellcheck, and it doesn't remind me to quote "`$generate`" as soon as I declare it with "`-i`"

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: you're right!
Maybe this is why tests pass!

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: (see above)

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: If `(( seen_leap++ ))` instead of `(( ++ seen_leap ))`, I got this error:

```
✗ random birthdates -> years are not leap years
(in test file baffling_birthdays.bats, line 111)
'(( seen_leap++ ))' failed
```
and I don't understand why

resu-xuniL:
It passed 4 times in a row before i push this, and now it failed for 10 and 23 people tests!
shall i increase back the tolerence (i decreased them)? or increase the runs to 3k ?

(in the php track, I ran it 10k ... but in bash it would take so much time !!)

resu-xuniL:
re-increased tolerance for `23 people` test.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
IsaacG: It's good style and good practice. It makes things consistent. It saves people from needing to look for a `declare` to know if there's missing quotes or not. There's no downside to it.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
IsaacG: Interesting. That sounds like `bats` runs with `set -e`. `v=0; (( v++ ))` evaluates `v` first then increments it (post-increment). `(( 0 ))` is false so the command returns non-zero which `set -e` flags. `(( ++v ))` pre-increments so `v=0; (( ++v ))` exits 0 (though `v=-1; (( ++v ))` would fail).

```suggestion
            (( ++seen_leap ))
```

IsaacG:
> (in the php track, I ran it 10k ... but in bash it would take so much time !!)

How much time is too much time? How slow is 3k? 4k?

If it takes more than, say, 5 seconds, we probably want to widen the tolerance.

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,70 @@
+#!/usr/bin/env bash
+
+main() {
+    local function=$1; shift
+
+    $function "$@"
```
IsaacG:
```suggestion
    "$function" "$@"
```

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,70 @@
...
+    local -i group_size=$1
+    # the more you increase the 'runs' value,
+    # the more the result is approching the expected one
+    # but the more you increase the computing time!
+    local -i runs=5000
+    local -i count=0
+
+    for ((i = 0; i <= runs; i++)); do
+        if [[ $(shared_birthday $(random_birthdates $group_size)) == "true" ]]; then
+            (( ++ count ))
```
IsaacG:
```suggestion
        if [[ "$(shared_birthday "$(random_birthdates "$group_size")")" == "true" ]]; then
            (( ++count ))
```

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,70 @@
...
+        fi
+    done
+
+    bc <<< "$count * 100 / $runs"
```
IsaacG: Do you need `bc` here or is an int result sufficient?

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    generate=500
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        month=${date:5:2}
+
+        seen_month[$month]=true
+    done
+    assert_equal ${#seen_month[@]} 12
```
IsaacG:
```suggestion
    assert_equal "${#seen_month[@]}" 12
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    generate=500
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        day=${date:8}
+
+        seen_day[$day]=true
+    done
+    assert_equal ${#seen_day[@]} 31
```
IsaacG:
```suggestion
    assert_equal "${#seen_day[@]}" 31
```

resu-xuniL:
> > (in the php track, I ran it 10k ... but in bash it would take so much time !!)
> 
> How much time is too much time? How slow is 3k? 4k?
> 
> If it takes more than, say, 5 seconds, we probably want to widen the tolerance.

it was about 20+ minutes for 2k - 3k before modifications...

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_output "true"
+}
+
+@test "random birthdates -> generate requested number of birthdates" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=500
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    assert_equal ${#output_dates[@]} "$generate"
```
glennj: Alternately:
```suggestion
    num_output_dates=$( wc -l << "$output" )
    assert_equal "$num_output_dates" "$generate"
```

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: I have to keep
```if [[ "$(shared_birthday $(random_birthdates "$group_size"))" == "true" ]]; then```
or it doesnt work anymore (the arguments must be splitted)

File: exercises/practice/baffling-birthdays/baffling_birthdays.bats [hunk 20]
```diff
@@ -0,0 +1,178 @@
...
+
+        seen_day[$day]=true
+    done
+    assert_equal ${#seen_day[@]} 31
+}
+
+# "The expected probability values should be compared using some tolerance to allow for small deviations."
+@test "estimated probability of at least one shared birthday -> for one person" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="0.0" # this is the exercice ref.
```
glennj: That is a confusing double comment. Perhaps:
```suggestion
    # The expected probability is 0.0
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    assert_success
+    assert_equal "$output" 0
+}
+
+@test "estimated probability of at least one shared birthday -> among ten people" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    # expect="11.694818" # this is the exercice ref.
+    run bash baffling_birthdays.sh estimated_probability_of_share_birthday 10
+    assert_success
+    (( output >= 10 && output <= 13 ))
```
glennj: This might be an opportunity to add a custom assertion
```bash
assert_between 10 13
# or
assert_closeto --tolerance=1.5 11.694818
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    generate=500
+    seen_leap=0
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        year=${date:0:4}
+        if (( year % 400 == 0 || (year % 4 == 0 && year % 100 != 0) )); then
+            (( ++ seen_leap ))
```
glennj: echoing Isaac's suggestion
```suggestion
            (( ++seen_leap ))
```

glennj (changes requested):

glennj:
Regarding the slowness, you'll get must faster performance by using the builtin [`$RANDOM`](https://www.gnu.org/software/bash/manual/bash.html#index-RANDOM) variable instead of calling out to shuf
```bash
# return a random number in the range 0 <= n < $1
random_int() {
    echo $(( RANDOM % $1 ))
}
february_days=$(( 1 + $(random_int 28) ))
```
With group_size = 10 and runs = 5000, you're calling shuf 150,000 times and date 50,000 times.

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,70 @@
...
+            (( random_year ++ ))
+        fi
+
+        case "$random_month" in
+                   2) random_day=$(shuf -n 1 -i 1-28) ;;
+            4|6|9|11) random_day=$(shuf -n 1 -i 1-30) ;;
+                   *) random_day=$(shuf -n 1 -i 1-31) ;;
+        esac
+
+        birthdates+=( "$(date -d "${random_year}-${random_month}-${random_day}" +%F)" )
```
glennj: You don't need to call out to date just to format it
```suggestion
        birthdates+=( "$(printf '%d-%02d-%02d' "$random_year" "$random_month" "$random_day")" )
```

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,70 @@
...
+        echo "true"
+    fi
+}
+
+random_birthdates () {
+    local -i number=$1
+    local -i actual_year random_year random_month random_day
+    local -a birthdates
+
+    actual_year=$(date -d "now" +%Y)
```
glennj: This can be a global variable, so you don't need to call date once for each run.

exercises/practice/baffling-birthdays/.meta/example.sh
glennj: Do it like you do in the tests

```bash
read -ra birthdates < <(random_birthdates "$group_size")
if [[ "$(shared_birthday "${birthdates[@]}")" == "true" ]]; then ...
```

Note that `true` and `false` are valid bash commands, so it's valid to write

```bash
if "$(shared_birthday "${birthdates[@]}")"; then ...
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: Would this custom assertion to be created in the `bats-extra.bash` file?

glennj: Yes. We won't be able to see it by default, so leave a comment there as a bookmark.

resu-xuniL: add custom `assert_between` in the `bats-extra.bash` file

```
assert_between() {
  local -r start="${1}"
  local -r end="${2}"

  if (( start > end )); then
    return 1
  elif ! (( output >= "$1" && output <= "$2" )); then
    batslib_print_kv_single_or_multi 6 \
      'start'  "${start}" \
      'end'    "${end}" \
      'output' "${output}" \
    | batslib_decorate 'output falls outside the range defined by start and end' \
    | fail
  fi
}
```

resu-xuniL: if fails this would gives:

```
   (from function `assert_between' in file bats-extra.bash, line 651,
    in test file baffling_birthdays.bats, line 162)
     `assert_between 10 13' failed
   
   -- output falls outside the range defined by start and end --
   start  : 10
   end    : 13
   output : 14
   --

```

resu-xuniL:
https://github.com/exercism/bash/pull/779/commits/bf5b0d01b02c57f360063b8f4e20296390fa613e

```
assert_between() {
  local -r start="${1}"
  local -r end="${2}"

  if (( start > end )); then
    return 1
  elif ! (( output >= "$1" && output <= "$2" )); then
    batslib_print_kv_single_or_multi 6 \
      'start'  "${start}" \
      'end'    "${end}" \
      'output' "${output}" \
    | batslib_decorate 'output falls outside the range defined by start and end' \
    | fail
  fi
}
```
on fail:
```
   (from function `assert_between' in file bats-extra.bash, line 651,
    in test file baffling_birthdays.bats, line 162)
     `assert_between 10 13' failed
   
   -- output falls outside the range defined by start and end --
   start  : 10
   end    : 13
   output : 14
   --
```

glennj:
> OK, it passes 6 times in a row (with this one) so far... Thank you for your help! (with runs at 1500)

The "random" exercises are like that. Don't forget, once this PR is merged, it will also run for every subsequent merge (one of the GHA workflows tests all exercises). So we'd want it succeeding just about every time.

exercises/practice/baffling-birthdays/bats-extra.bash
```diff
@@ -635,3 +635,19 @@ refute_output() {
...
+  local -r end="${2}"
+
+  if (( start > end )); then
+    return 1
+  elif ! (( output >= "$1" && output <= "$2" )); then
+    batslib_print_kv_single_or_multi 6 \
+      'start'  "${start}" \
+      'end'    "${end}" \
+      'output' "${output}" \
+    | batslib_decorate 'output falls outside the range defined by start and end' \
```
glennj:
```suggestion
    | batslib_decorate 'output falls outside the inclusive range defined by start and end' \
```

exercises/practice/baffling-birthdays/bats-extra.bash
```diff
@@ -635,3 +635,19 @@ refute_output() {
...
+  local -r start="${1}"
+  local -r end="${2}"
+
+  if (( start > end )); then
+    return 1
+  elif ! (( output >= "$1" && output <= "$2" )); then
+    batslib_print_kv_single_or_multi 6 \
+      'start'  "${start}" \
+      'end'    "${end}" \
+      'output' "${output}" \
```
glennj:
```suggestion
      'actual' "${output}" \
```

exercises/practice/baffling-birthdays/bats-extra.bash
```diff
@@ -635,3 +635,19 @@ refute_output() {
...
+    | batslib_decorate 'output falls outside the range defined by start and end' \
+    | fail
+  fi
+}
```
glennj: last line is missing the trailing newline. Please update your editor settings to add it if it's missing.

glennj (changes requested):

exercises/practice/baffling-birthdays/bats-extra.bash
resu-xuniL: I use to use a `.editorconfig' file, but I had to remove it not to include it in the PR.
Just forgot to use it again before PR.

Won't you add on in the project root dir, whith your specifications?

```
# EditorConfig is awesome: https://EditorConfig.org

root = true

###########################################################
; common
###########################################################

[*]
charset = utf-8

end_of_line = lf
insert_final_newline = true
trim_trailing_whitespace = true

indent_style = space
indent_size = 4

###########################################################
; markdown
###########################################################

[*.md]
trim_trailing_whitespace = false

```

resu-xuniL:
> > OK, it passes 6 times in a row (with this one) so far... Thank you for your help! (with runs at 1500)
> 
> The "random" exercises are like that. Don't forget, once this PR is merged, it will also run for every subsequent merge (one of the GHA workflows tests all exercises). So we'd want it succeeding just about every time.

I agree. I'll run many tests on 2 machines to see if it fails, and if it does, maybe we would increase the tolerance?

so far, 8 in a row with this one
From now on, I'm going to run the `10` and `23` people tests only. These were the only ones that were failing.
(and I'll won't waste (big) time with the `70` people test!)


**Edit1**: 25 runs and no fails.... 
26th fail : output 47 (for range 48-53)

stats:
```
output for test `10 people` : 
(assert_between is 10-13)
6x 10 | 13x 11 | 7x 12 | 0x 13 
none 13. Shall we 'risk' to decrease to 12?
```
```
output for test `23 people` :
(assert_between is 48-53)
2x 48 | 4x49 | 13x50 | 5x51 | 0x 52 | 1x 53 
+ 1x 47 !!! --> need to increase to 47-53
```

**Edit2**: I'll run with 3k runs (instead of 2k) to see results (with ranges: `10-12` and `48-52`)
another fail at 53. need to keep 53

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,75 @@
...
+
+    if (( ${#birthdays[@]} == ${#birthdates[@]} )); then
+        echo "false"
+    else
+        echo "true"
+    fi
+}
+
+random_int() {
+    echo $(( RANDOM % $1 ))
```
IsaacG: Unquoted expansion

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,75 @@
...
+    local -i number=$1
+    local -i random_year random_month random_day
+    local -a birthdates
+
+    for (( i = 0; i < number; i++ )); do
+        random_year=$(( 1900 + $(random_int $(( current_year - 1900 + 1 ))) ))
+        random_month=$(( 1 + $(random_int 12) ))
+
+        if (( random_year % 400 == 0 || (random_year % 4 == 0 && random_year % 100 != 0) )); then
+            (( random_year ++ ))
```
IsaacG:
```suggestion
            (( random_year++ ))
```

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,75 @@
...
+        fi
+
+        case "$random_month" in
+                   2) random_day=$(( 1 + $(random_int 28) )) ;;
+            4|6|9|11) random_day=$(( 1 + $(random_int 30) )) ;;
+                   *) random_day=$(( 1 + $(random_int 31) )) ;;
+        esac
+
+        birthdates+=( "$(printf '%d-%02d-%02d' "$random_year" "$random_month" "$random_day")" )
```
IsaacG: Note, `printf` has a `-v` to set vars.

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,75 @@
...
+    local -i group_size=$1
+    # the more you increase the 'runs' value,
+    # the more the result is approching the expected one
+    # but the more you increase the computing time!
+    local -i runs=2000
+    local -i count=0
+
+    for (( i = 0; i <= runs; i++ )); do
+        read -ra birthdates < <(random_birthdates "$group_size")
+        if [[ "$(shared_birthday "${birthdates[@]}")" == "true" ]]; then
```
IsaacG: Optionally you can directly evaluate true/false:
```suggestion
        if "$(shared_birthday "${birthdates[@]}")"; then
```

exercises/practice/baffling-birthdays/.meta/example.sh
```diff
@@ -0,0 +1,75 @@
...
+            (( count++ ))
+        fi
+    done
+
+    echo $(( count * 100 / runs ))
```
IsaacG: Unquoted expansion

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+        if (( year % 400 == 0 || (year % 4 == 0 && year % 100 != 0) )); then
+            (( ++seen_leap ))
+        fi
+    done
+    assert_equal "${seen_leap}" 0
+}
+
+@test "random birthdates -> months are random" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    declare -A seen_month
```
IsaacG, line 119: You can use a regular array here:

```suggestion
    declare -a seen_month
```

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    generate=500
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        month=${date:5:2}
+
+        seen_month[$month]=true
+    done
+    assert_equal "${#seen_month[@]}" 12
```
IsaacG, line 129: Should we test for a semi-uniform distribution? Each month should have 20-80 birthdays.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
```diff
@@ -0,0 +1,178 @@
...
+    generate=500
+    run bash baffling_birthdays.sh random_birthdates "$generate"
+    assert_success
+    read -ra output_dates <<< "$output"
+    for date in "${output_dates[@]}"; do
+        day=${date:8}
+
+        seen_day[$day]=true
+    done
+    assert_equal "${#seen_day[@]}" 31
```
IsaacG, line 144: Check for a uniform distribution (excepting 31)?

exercises/practice/baffling-birthdays/.meta/example.sh
resu-xuniL: another one! (need to be more carefull)

exercises/practice/baffling-birthdays/bats-extra.bash
glennj: Looking at editorconfig.org, it seems like you could put it in the parent directory of your bash fork. Alternatively, for personal editor files, use your personal gitignore file (~/.config/git/ignore)

However, I wouldn't mind putting this in the repo. @IsaacG ?

resu-xuniL: > (~/.config/git/ignore)

didn't knew this. thanks !

exercises/practice/baffling-birthdays/baffling_birthdays.bats
glennj: Wouldn't that be testing bash's randomness, not the student's solution?

IsaacG: Yes and no. This would guard against a solution that adds one birthdate to each month then puts the rest all in January, for example. A perfectly uniform distribution would have roughly `500/12 = 41` per month. Keeping a wide band (eg 20-60) would allow for randomness. We _do_ want to test that the randomness is somewhat uniform, both in the student's code and with bash's randomness. If the randomness generator used always returns `4`, then we want to fail the solution ... even if that means we're testing the randomness.

exercises/practice/baffling-birthdays/bats-extra.bash
IsaacG: https://forum.exercism.org/t/proposal-add-editorconfig-for-consistent-formatting/18593

https://discord.com/channels/854117591135027261/1398672004059500648 (private channel)

I'd be open to it but I think it makes more sense to try to get this into the generic track first (as something explicitly marked as optional, to be decided by track maintainers) with some common consensus on good values, then pull it from there.

exercises/practice/baffling-birthdays/baffling_birthdays.bats
glennj: If that happens upstream then sure, but I don't think it's important now. We can never prevent students from cheating entirely.

exercises/practice/baffling-birthdays/bats-extra.bash
glennj: (I unresolved this just to surface this conversation: I clicked away from it then had trouble finding it again)

I can't imagine that we could possibly agree on many editorconfig settings in a generic track. But this discussion is no longer about this PR.

resu-xuniL:
Feedback: after multiple tests (with `runs` set at `2k` or `3k`), the range output is `10-12` and `48-51` most of the times, **BUT** for some randomness reason, it sometimes gives very few `13`, `47` or `53` (and this is very very rare. It even gave once a `9`!).

The safer range would then be `9-13` and `47-53` (which is a big tolerance range).

I'm testing it now with runs set at `5k` (test is now 1/2 hour to complete).

resu-xuniL:
Results so far (runs are set at `5k`):
```
11 50 99
11 51 99
11 50 99
10 50 99
11 51 99
12 52 99
11 50 99
11 51 100
12 50 99
11 49 99
11 50 100
11 50 100
11 50 99
```
very rare are the `9` or the `13` 
very rare are the `49` or the `52` 
never seen is the `98`

--

To prevent the failure, maybe we could add this condition in the .bats file (this would be **very** unlucky if the run fails twice in a row!)

```
@test "estimated probability of at least one shared birthday -> among ten people" {
    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
    # The expected probability is 11.694818
    run bash baffling_birthdays.sh estimated_probability_of_share_birthday 10
    assert_success
    if ! assert_between 10 12; then
        run bash baffling_birthdays.sh estimated_probability_of_share_birthday 10
        assert_success
        assert_between 10 12
    fi
}

@test "estimated probability of at least one shared birthday -> among twenty-three people" {
    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
    # The expected probability is 50.729723
    run bash baffling_birthdays.sh estimated_probability_of_share_birthday 23
    assert_success
    if ! assert_between 49 51; then
        run bash baffling_birthdays.sh estimated_probability_of_share_birthday 23
        assert_success
        assert_between 49 51
    fi
}

@test "estimated probability of at least one shared birthday -> among seventy people" {
    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
    # The expected probability is 99.915958
    run bash baffling_birthdays.sh estimated_probability_of_share_birthday 70
    assert_success
    if ! assert_between 98 100; then
        run bash baffling_birthdays.sh estimated_probability_of_share_birthday 70
        assert_success
        assert_between 98 100
    fi
}
```
I'm testing this ATM, and it seems to work good.

IsaacG:
Note, the tests need to run in under 20s max. I'd recommend trying to get them in under 5s.

resu-xuniL:
well, then, I can't see how to do... unless removing the  tests which check for "estimated probability of at least one shared birthday"

IsaacG:
This is a bit of a hacky solution, but it is valid and fast. Roughly 0.9s to run 5000 checks, each check looking for a collision with up to 70 random values.

```bash
#!/bin/bash

check () {
    local -a hit
    for (( j = 0; j < $1; j++ )); do
        day=$((RANDOM % 365))
        (( hit[day]++ == 0 )) || return 1
    done
}

for (( i = 0; i < 5000; i++ )); do
    check 70 || (( hit++ ))
done

echo $hit $(( hit / 50 ))
```

It bypasses the whole "generate 70 dates" and goes straight to the math side: generate 70 values from a space of 365 and look for a collision.

resu-xuniL:
edit: Isn't the point to test the whole "generated 70 dates" ?

glennj:
I conclude that bash cannot be used to solve this with nested loops. It's just too slow.

You might want to investigate  https://en.wikipedia.org/wiki/Birthday_problem#Approximations

glennj:
I had a conversation with Claude about it. Turns out the bottlenecks are the number of subshells. (The "200 subshells" you see are while I had reduced `runs=2000`)

> # Summary
> 
> The performance bottleneck in the original `estimated_probability_of_share_birthday` was **massive subshell spawning** — 200 iterations × multiple subprocess forks per iteration = thousands of process forks.
> 
> ## Root Causes Eliminated
> 
> | Problem | Original | Optimized |
> | --- | --- | --- |
> | `random_int` helper | `$(random_int N)` — subshell per call | Inlined: `RANDOM % N` directly in arithmetic |
> | `printf` formatting | `$(printf '%d-%02d-%02d' ...)` — subshell per date | Manual zero-padding with `&&`/`\|\|` in-process (Glenn: used `printf -v`) |
> | `random_birthdates` call | Process substitution `< <(random_birthdates ...)` — 200 subshells | Inlined loop via `_make_random_birthdate` (sets global `_random_birthday`) |
> | `shared_birthday` call | Command substitution `"$(shared_birthday ...)"` — 200 subshells | Inlined check via `_check_shared_birthday` (sets global `_has_shared`) |
> 
> ## Design Pattern
> 
> Two helper functions use **output globals** instead of stdout to avoid subprocess forks:
> 
> * `_make_random_birthdate` → sets `$_random_birthday`
> * `_check_shared_birthday` → sets `$_has_shared`
> 
> The public `random_birthdates` and `shared_birthday` functions remain unchanged for compatibility with other test cases.

Here's a version that uses "private" functions to write global variables as the data-passing mechanism: https://gist.github.com/glennj/4e8a3ab4a9942b798ace0f8dba0bb30b

The test suite runs in about 7 seconds on my machine.

IsaacG:
If this exercise requires hoops to solve in under 5s, I propose we add it to the foregone list.

resu-xuniL:
I don't think this is 'hoops', this is just that I need learning more and more :)

resu-xuniL:
new stats with runs=2k
```
      1 x 9
     12 x 10
     65 x 11
     40 x 12
      8 x 13

      1 x 47
      5 x 48
     15 x 49
     28 x 50
     25 x 51
     11 x 52
      3 x 53
```

IsaacG:
The fact  that this requires so much work and learning to get working in a reasonable amount of time indicates to me that this may not be a great fit for the bash track :smile: 

Thoughts on marking this foregone, @glennj ?

glennj:
> The fact that this requires so much work and learning to get working in a reasonable amount of time indicates to me that this may not be a great fit for the bash track 😄
> 
> Thoughts on marking this foregone, @glennj ?

We have:

1. a monte-carlo approach that takes an unreasonable amount of time,
1. a monte-carlo approach that is reasonable, and
1. non-randomized mathematical approaches (not demonstrated here, but https://exercism.org/tracks/lua/exercises/baffling-birthdays/solutions has some)

I think we can proceed. We'll probably see monstly 1 and 3 submitted.

An instructions.append.md file would be useful to gently steer people along. Approaches articles might help too.

IsaacG:
> I think we can proceed. We'll probably see monstly 1 and 3 submitted.
> 
> An instructions.append.md file would be useful to gently steer people along. Approaches articles might help too.

Solutions that take long to run will fail the test runner. The approaches article can only be viewed after passing tests.

We'd want to provide a fair bit of guidance in the append file to help students get this to pass on the test runner.

glennj:
How about this approach:

1. add `"status": "wip"` to the exercise blob in config.json and we merge this PR.
2. then we work on a new PR to add `.docs/instructions.append.md` and change the status to `"beta"`.

IsaacG:
What's the status on this PR?

config.json
```diff
@@ -655,6 +655,17 @@
...
+      {
+        "slug": "baffling-birthdays",
+        "name": "Baffling Birthdays",
+        "uuid": "c3fbce6a-61ce-439c-8c67-fad93561326f",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 4,
+        "topics": [
+          "dates"
+        ]
```
glennj, line 667: Please set the exercise status as previously requested.
```suggestion
        ],
        "status": "wip"
```

glennj (changes requested):

glennj:
@IsaacG I think this is ready to merge. The exercise status is "wip".

glennj (approved):

exercises/practice/baffling-birthdays/baffling_birthdays.bats
IsaacG, line 1: Should this be template-generated?

IsaacG (approved):

exercises/practice/baffling-birthdays/baffling_birthdays.bats
glennj: I would say no. There are some template-able tests at the start and at the end, but there's a big blob in the middle that are one-offs. From templating in tcl and moonscript I'd say don't bother.

resu-xuniL:
fixes at #893

exercises/practice/baffling-birthdays/baffling_birthdays.bats
resu-xuniL: OK

Issue #783, opened 2026-06-15
Add github action to shellcheck bash repo


glennj:
This is to ensure code quality.

Inspired by shellcheck usage in factor-test-runner.

Todo:

- [ ] Fix any findings in bats files
- [ ] Fix any findings in example solutions
- [ ] Add shellcheck.yaml to .github/workflows

glennj:
I'll exclude

* check [SC1091](https://github.com/koalaman/shellcheck/wiki/SC1091) ("not following 'source somefile'")
* single file `exercises/practice/markdown/markdown.sh`, the purposefully bad file to refactor

PR #784, opened 2026-06-15
Shellcheck part 1: bin scripts and non-example exercise files

glennj:
Ref: issue #783 

This PR will satisfy

```sh
find exercises/practice bin \
    \( -name '*.sh' -o -name '*.bash' \) \
    -not \( -name example.sh -o -path exercises/practice/markdown/markdown.sh \) \
    -print0 \
| xargs -0 shellcheck
```

I'd recommend reviewing by commit: the last commit is just copying one bats-extra.bash file to all the rest of the exercises.

### DIFF
--- a/exercises/practice/acronym/bats-extra.bash
+++ b/exercises/practice/acronym/bats-extra.bash
@@ -20,7 +20,9 @@
 # <http://creativecommons.org/publicdomain/zero/1.0/>.
 #
 
+# shellcheck disable=SC2120 # (warning): fail references arguments, but none are ever passed.
 fail() {
+  # shellcheck disable=2015 # (info): Note that A && B || C is not if-then-else. C may run when A is true.
   (( $# == 0 )) && batslib_err || batslib_err "$@"
   return 1
 }
@@ -119,7 +121,7 @@ batslib_print_kv_single_or_multi() {
   else
     local -i i
     for (( i=1; i < ${#pairs[@]}; i+=2 )); do
-      pairs[$i]="$( batslib_prefix < <(printf '%s' "${pairs[$i]}") )"
+      pairs[i]="$( batslib_prefix < <(printf '%s' "${pairs[$i]}") )"
     done
     batslib_print_kv_multi "${pairs[@]}"
   fi
@@ -136,6 +138,7 @@ batslib_prefix() {
 batslib_mark() {
   local -r symbol="$1"; shift
   # Sort line numbers.
+  # shellcheck disable=SC2046 # (warning): Quote this to prevent word splitting.
   set -- $( sort -nu <<< "$( printf '%d\n' "$@" )" )
 
   local line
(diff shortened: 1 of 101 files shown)


exercises/practice/list-ops/list_ops.bats
```diff
@@ -15,35 +17,35 @@ setup() { source list_ops.sh; }
 
 @test "append empty lists" {
     #[[ $BATS_RUN_SKIPPED == "true" ]] || skip
-    list1=()
-    list2=()
+    local list1=()
+    local list2=()
```
glennj, line 21: The test variables are now localized: shell check was complaining about `result` being used as a scalar after previously being an array.

exercises/practice/acronym/bats-extra.bash
```diff
@@ -366,6 +369,8 @@ assert_output() {
       | fail
     fi
   elif (( is_mode_regexp )); then
+    # glennj: I don't know what command's status `$?` is at this point
+    # shellcheck disable=SC2319 # (warning): This $? refers to a condition, not a command. Assign to a variable to avoid it being overwritten.
```
IsaacG: I'm pretty sure `$?` is set by `[[ '' =~ $expected ]]`. From `man bash`,

> An additional binary operator, =~, is available [..].  If the regular expression is syntactically incorrect, the conditional expression's return value is 2.

This attempts to use `$expected` as a regex and checks if it is an invalid regex.

IsaacG (approved):

exercises/practice/acronym/bats-extra.bash
```diff
@@ -369,9 +369,9 @@ assert_output() {
       | fail
     fi
   elif (( is_mode_regexp )); then
-    # glennj: I don't know what command's status `$?` is at this point
-    # shellcheck disable=SC2319 # (warning): This $? refers to a condition, not a command. Assign to a variable to avoid it being overwritten.
+    # shellcheck disable=SC2319 # (warning) `$?` in a conditional expression: https://github.com/koalaman/shellcheck/wiki/SC2319
     if [[ '' =~ $expected ]] || (( $? == 2 )); then
+      # the regex matches an empty string or it is syntactically incorrect.
```
IsaacG, line 374: I suspect matching the empty string is not intentional. We could allow an empty match and/or fail explicitly on an empty match.

```
[[ '' =~ $expected ]]
match=$?
msg=""
if (( match == 0 )); then
  msg="Regex matches an empty string"
elif (( match == 2 )); then
  msg="Invalid regex"
fi
if [[ -n $msg ]]; then
  echo "${msg}: ${expected@Q}"
  ...

glennj: > I suspect matching the empty string is not intentional.

Betcha it is.

I don't really want to diverge the code we took from bats-core/bats-assert.

IsaacG: We could add an ignore for these files. There's an ignore option in the workflow.

PR #844, opened 2026-06-29
[house] Add a Jinja test template

IsaacG:
This changes the expecations for the output layout.
This brings the exercise in line with the problem spec and breaks all existing solutions.

### DIFF
--- a/exercises/practice/house/house.bats
+++ b/exercises/practice/house/house.bats
@@ -1,22 +1,23 @@
 #!/usr/bin/env bats
 load bats-extra
 
-# local version: 2.2.0.1
-# bash-specific test: Input validation
+# generated on 2026-06-29T18:35:03+00:00
 
-@test "verse 1" {
-    #[[ $BATS_RUN_SKIPPED == "true" ]] || skip
-    expected="This is the house that Jack built."
+@test "verse one - the house that jack built" {
+    # [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    expected=$(cat <<END
+This is the house that Jack built.
+END
+)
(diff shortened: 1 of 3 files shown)


kotp (approved):

IsaacG:
+cc @exercism/bash  This changes the expected output format to match the problem spec, breaking existing implementations. Thoughts on changing this exercise in that way? Any objections?

Personally, I think moving things towards the problem spec is generally good.

kotp:
I am for it, and the output change is likely an easy change for the solutions that exist.  We could still commit with the "[no important files changed]" if we want to have students to opt-in on the changes, without invalidating their solutions, right?

exercises/practice/house/house.bats
```diff
@@ -1,348 +1,178 @@
 #!/usr/bin/env bats
 load bats-extra
 
-# local version: 2.2.0.1
-# bash-specific test: Input validation
+# generated on 2026-06-29T04:05:11+00:00
+# local version: 2.0.0.0
```
glennj: Does anyone remember why we started putting version numbers into the tests? This change is going backwards.

IsaacG: I have no idea why we do those. I'm in favor of dropping them.

Do you have preferences on adding the test UUIDs for each test (could be done at the same time)?

glennj: We don't currently have a reason to include the UUIDs. Unless we want to use them to disambiguate tests with the same name in different test groups.

IsaacG: Done

glennj (approved):

PR #850, opened 2026-06-29
[ocr-numbers] Add a Jinja test template

IsaacG:


kotp (approved):

exercises/practice/ocr-numbers/.meta/template.j2
```diff
@@ -0,0 +1,24 @@
...
+}
+{% endfor %}
+
+@test "No input" {
+    [[ $BATS_RUN_SKIPPED == "true" ]] || skip
+    run bash {{ solution }}
+    assert_success
+    assert_output ""
+}
+
```
kotp: The extra newline/line termination should not be here from what I can tell.

IsaacG: Removed. Thanks!

PR #853, opened 2026-06-29
[perfect-numbers] Add a Jinja test template

IsaacG:
This updates the error message to match the problem specs, breaking existing solutions.

### DIFF
--- a/exercises/practice/perfect-numbers/.meta/example.sh
+++ b/exercises/practice/perfect-numbers/.meta/example.sh
@@ -29,7 +29,7 @@ aliquot_sum() {
 declare -i num=$1
 
 if ((num <= 0)); then
-    echo "Classification is only possible for natural numbers." >&2
+    echo "Classification is only possible for positive integers." >&2
     exit 1
 fi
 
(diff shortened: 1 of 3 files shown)


kotp (approved):

exercises/practice/perfect-numbers/.meta/example.sh
```diff
@@ -29,7 +29,7 @@ aliquot_sum() {
 declare -i num=$1
 
 if ((num <= 0)); then
-    echo "Classification is only possible for natural numbers." >&2
+    echo "Classification is only possible for positive numbers." >&2
```
kotp: The wording changed here, and the tests are failing in the relevant place... since natural number and 0 are not mutually exclusive.

kotp (commented):
Is the solution file, the example file, that no longer is correct for the change in the generated test?

exercises/practice/perfect-numbers/.meta/example.sh
IsaacG: updated

IsaacG:
> Is the solution file, the example file, that no longer is correct for the change in the generated test?

bash had "natural numbers". The problem spec has "positive integers". I accidentally changed the example solution from "natural numbers" to "positive numbers", correcting one of the two words.

The example is now in sync with the problem specs.

kotp (approved):
Looking good, passing all tests.

PR #856, opened 2026-06-29
[pythagorean-triplet] Add a Jinja test template

IsaacG:


kotp (approved):

kotp (commented):
Error message states it... "should be exactly 2 commented skip directive (sic) in pythagorean_triplet.bats"

Might also look at the plural/singular slip there as well.

IsaacG:
> Error message states it... "should be exactly 2 commented skip directive (sic) in pythagorean_triplet.bats"
> 
> Might also look at the plural/singular slip there as well.

Updated

kotp:
This (The tests, I mean) is running "forever" for me locally.

```
Processing pythagorean-triplet  
pythagorean_triplet.bats        
 ✓ triplets whose sum is 12     
 ✓ triplets whose sum is 108    
 ✓ triplets whose sum is 1000   
 ✓ no matching triplets for 1001
 ✓ returns all matching triplets
 ✓ several matching triplets    
   triplets for large number 
```

IsaacG:
> This (The tests, I mean) is running "forever" for me locally.

I changed the skip-logic in the template but forget to regenerate the test file. It should be better now.

The large test is meant to be skipped.

PR #884, opened 2026-07-18
[swift-scheduling] Sync new test from the problem specs

IsaacG:


IsaacG:
This new test may break some implementations

kotp (approved):

glennj (approved):

